Issue #72, opened 2013-07-27
Object Prototypes - extending objects at runtime


Ocramius:
Using the access interceptor logic (or similar logic) it should be quite easy to implement object prototypes.

An example DSL for object prototypes in PHP would be like following:

``` php
class Foo
{
    public function doFoo()
    {
        return 'foo';
    }

    public function doBar()
    {
        return 'bar';
    }
}

$object = $prototypeApplier->applyPrototype(
    new Foo(),
    array(
        'doFoo' => function () { return 'baz'; },
        'addedMethod' => function () { return 'new method'; }
    )
);

echo $object->doFoo() . "\n"; // 'baz'
echo $object->doBar() . "\n"; // 'bar'
echo $object->addedMethod() . "\n"; // 'new method'
```

Additionally, a prototype may also implement some additional interfaces on the wrapper at runtime, which probably calls for a `PrototypeDefinition` class or builder. This can be useful when we want to extend an API and still keep type-safety:

PHP 5.3 could be supported via code generation and something like [super_closure](https://github.com/jeremeamia/super_closure), but I'm not sure if it's worth it

Ocramius:
See discussion in #103 about why I closed this. Great idea, but too much maintenance burden with the current development approach.

PR #103, opened 2013-10-02
[TO BE MOVED] Parametric polymorphism proxy / Object method extensions

blanchonvincent:
Interfaces are supported now.

https://github.com/Ocramius/ProxyManager/issues/72 is done with this PR
- [x] unit tests
- [x] integration tests
- [x] documentation
- [x] examples
- [x] documentation generator

Waiting for (merged in 2.3.0) :
- [x] https://github.com/zendframework/zf2/pull/5245 (to replace ReflectionTools)
- [x] https://github.com/zendframework/zf2/pull/5262 (to use in documentation)


tklever:
Did this ever get moved / landed somewhere? #72 is the exact problem I'm looking to solve (extra interfaces at runtime)

Ocramius:
I never managed to continue working on it. The idea is very interesting and useful for [DCI](https://en.wikipedia.org/wiki/Data,_context_and_interaction), but we're far from making it a live project.

Specifically, I think we'd need some sort of way to reuse most of the codegen tools inside ProxyManager to make it work.

Ocramius:
This is something that I won't be implementing on this library anymore (simply because of the added maintenance complexity), but new tools have arisen to allow for it to work properly.

Specifically, I'm thinking of https://github.com/Roave/BetterReflection, which allows runtime monkey-patching. I can probably get back at this once Ocramius/ProxyManager once the proxymanager is fully AST-based (no more zend-code), but we're still far from getting there.

Meanwhile, closing as `wontfix`

Issue #210, opened 2014-12-16
HHVM compatibility


Ocramius:
- [x] current `master` fatals because of the new `GhostObject` implementation based on `Closure::bind()` awesomeness
- [x] facebook/hhvm#4465
- [ ] facebook/hhvm#4560

Ocramius:
HHVM still fails in numerous ways. It works for basic use-case scenarios, but we still have issues with:
 * private properties with the same name inside an inheritance
 * variadic arguments
 * by-ref variadic arguments
 * typed variadic arguments

Ocramius:
This is not going to happen anymore, since the ecosystem is slowly rejecting (sadly, but reasonably) HHVM's long incompatibility list

PR #227, opened 2015-01-18
Temporary workaround for #210 HHVM compatibility (see facebook/hhvm#4560)

Ocramius:
See #210

This PR breaks private/protected properties encapsulation in HHVM, but it is an acceptable tradeoff until facebook/hhvm#4560 is fixed


Ocramius:
Fix still seems to be invalid for now.

Ocramius:
This is just wrong. Closing.

Issue #226, opened 2015-01-14
Avoid duplication on Generators class


malukenho:
As we know, today we have code duplication in the configuration of classes "Generators".
```php
                [
                    $init,
                    new StaticProxyConstructor(
                        $initializer,
                        Properties::fromReflectionClass($originalClass)
                    ),
                    new MagicGet(
                        $originalClass,
                        $initializer,
                        $init,
                        $publicProperties,
                        $protectedProperties,
                        $privateProperties
                    ),
                    new MagicSet(
                        $originalClass,
                        $initializer,
                        $init,
                        $publicProperties,
                        $protectedProperties,
                        $privateProperties
                    ),
                    new MagicIsset(
                        $originalClass,
                        $initializer,
                        $init,
                        $publicProperties,
                        $protectedProperties,
                        $privateProperties
                    ),
                    new MagicUnset(
                        $originalClass,
                        $initializer,
                        $init,
                        $publicProperties,
                        $protectedProperties,
                        $privateProperties
                    ),
                    new MagicClone($originalClass, $initializer, $init, $publicProperties),
                    new MagicSleep($originalClass, $initializer, $init, $publicProperties),
                    new SetProxyInitializer($initializer),
                    new GetProxyInitializer($initializer),
                    new InitializeProxy($initializer, $init),
                    new IsProxyInitialized($initializer),
                ]
```

I thinking in anything as a factory to generate values by values passed previously as follow example:
```php
$generatorFactory = new GeneratorFactory(
    $originalClass,
    $initializer,
    $init,
    $publicProperties,
    $protectedProperties,
    $privateProperties
);

$generatorFactory->addType('MagicClone')
    ->addType('MagicSleep')
    ->addType('IsProxyInitialized');

// Array with objects generated
$methods = $generatorFactory->asArrayList();
```

ping @Ocramius

Ocramius:
@malukenho I'm thinking of `GeneratorContext` instead, and adding specific getters to it. Thoughts?

Ocramius:
Absolutely no setters, that's also something that I want to make clear.

malukenho:
@Ocramius At the end there is the same? Has any advantage in `get` approach?

Ocramius:
@malukenho I'm not sure if there's an advantage, to be honest.

malukenho:
OK @Ocramius.
But I thinks Its not avoid the duplication.

```php
$generatorFactory = new GeneratorContext(
    $originalClass,
[...]

PR #230, opened 2015-02-04
Avoiding duplicate method generator

malukenho:
As spoken in issue #226, begin to create a mechanism to avoid duplication of code, especially the creation of complicated objects, with many parameters, such as generators methods. 

The idea is to centralize the changes in these methods, and facilitate the creation of multiple methods.


malukenho:
As discussed with @Ocramius on chat, hide the problem methods, not solve it. Also, the implementation of a factory, should be immutable, stateless. I'm closing this PR, keeping in mind that I have not found a good solution to actually solve the problem without hide it.

PR #233, opened 2015-03-19
Fix html titles

malukenho:


Ocramius:
@malukenho you didn't start off `master`

PR #237, opened 2015-04-01
remove unused title table at README.md ?

samsonasik:


Ocramius:
This is actually used by couscous to generate the HTML titles. It indeed looks bad, so I'll merge this.

/cc @malukenho

PR #243, opened 2015-07-13
Experiment: switching @asgrim's much better roave/better-reflection stuff

Ocramius:
Note: still incomplete, as roave/better-reflection didn't reach parity with internals reflection yet.

### DIFF
--- a/src/ProxyManager/Factory/AbstractBaseFactory.php
+++ b/src/ProxyManager/Factory/AbstractBaseFactory.php
@@ -87,7 +87,7 @@ abstract class AbstractBaseFactory
         $this
             ->configuration
             ->getSignatureChecker()
-            ->checkSignature(new ReflectionClass($proxyClassName), $proxyParameters);
+            ->checkSignature(\BetterReflection\Reflection\ReflectionClass::createFromName($proxyClassName), $proxyParameters);
 
         return $this->checkedClasses[$className] = $proxyClassName;
     }
@@ -112,7 +112,7 @@ abstract class AbstractBaseFactory
         $className = $this->configuration->getClassNameInflector()->getUserClassName($className);
         $phpClass  = new ClassGenerator($proxyClassName);
 
-        $this->getGenerator()->generate(new ReflectionClass($className), $phpClass, $proxyOptions);
+        $this->getGenerator()->generate(\BetterReflection\Reflection\ReflectionClass::createFromName($className), $phpClass, $proxyOptions);
 
         $phpClass = $this->configuration->getClassSignatureGenerator()->addSignature($phpClass, $proxyParameters);
 
--- a/src/ProxyManager/Factory/AbstractBaseFactory.php
+++ b/src/ProxyManager/Factory/AbstractBaseFactory.php
@@ -18,10 +18,11 @@
 
 namespace ProxyManager\Factory;
 
+use BetterReflection\Reflection\Adapter\ReflectionClass;
+use BetterReflection\Reflection\ReflectionClass as BetterReflectionClass;
 use ProxyManager\Configuration;
 use ProxyManager\Generator\ClassGenerator;
 use ProxyManager\Version;
-use ReflectionClass;
 
 /**
  * Base factory common logic
@@ -87,7 +88,10 @@ abstract class AbstractBaseFactory
         $this
             ->configuration
             ->getSignatureChecker()
-            ->checkSignature(\BetterReflection\Reflection\ReflectionClass::createFromName($proxyClassName), $proxyParameters);
+            ->checkSignature(
+                new ReflectionClass(BetterReflectionClass::createFromName($proxyClassName)),
+                $proxyParameters
+            );
 
         return $this->checkedClasses[$className] = $proxyClassName;
     }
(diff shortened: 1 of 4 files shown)


src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
```diff
@@ -62,6 +62,8 @@ public function generate(ClassGenerator $classGenerator)
         file_put_contents($tmpFileName, "<?php\n\n" . $generatedCode);
         rename($tmpFileName, $fileName);
 
+        eval($generatedCode);
```
Ocramius: To be removed before merge

src/ProxyManager/Factory/AbstractBaseFactory.php
```diff
@@ -112,7 +112,7 @@ private function generateProxyClass($proxyClassName, $className, array $proxyPar
         $className = $this->configuration->getClassNameInflector()->getUserClassName($className);
         $phpClass  = new ClassGenerator($proxyClassName);
 
-        $this->getGenerator()->generate(new ReflectionClass($className), $phpClass, $proxyOptions);
+        $this->getGenerator()->generate(\BetterReflection\Reflection\ReflectionClass::createFromName($className), $phpClass, $proxyOptions);
```
Ocramius: Requires Roave/BetterReflection#18 to work properly

src/ProxyManager/Factory/AbstractBaseFactory.php
```diff
@@ -87,7 +87,7 @@ protected function generateProxy($className, array $proxyOptions = [])
         $this
             ->configuration
             ->getSignatureChecker()
-            ->checkSignature(new ReflectionClass($proxyClassName), $proxyParameters);
+            ->checkSignature(\BetterReflection\Reflection\ReflectionClass::createFromName($proxyClassName), $proxyParameters);
```
Ocramius: Requires Roave/BetterReflection#18 to work properly

Ocramius:
Not for 2.0.0: we will likely use this for later versions

lisachenko:
@Ocramius you should also give a try for goaop/parser-reflection library. It's compatible with internal reflection (something like 95% now, only trait adaptation is not implemented yet) And I will work on it for my own dirty needs ;)

Ocramius:
@lisachenko eh, I actively worked with @asgrim on better-reflection, following internal design and such, so while I would love to try your lib, it would probably be much more time consuming to do so instead of just using something that I know very well :-P

lisachenko:
Ok ) I can understand your decision, because we select different goals. I want my library to be as close as possible to the original versions of reflection, so no extra methods and functionality. This should be a basis for top-level libraries...

PS. Looks like some kind of competition will improve our projects ;)

src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
```diff
@@ -74,6 +74,8 @@ public function generate(ClassGenerator $classGenerator)
             restore_error_handler();
         }
 
+        eval($generatedCode);
```
Ocramius: to be removed

Ocramius:
> Ok ) I can understand your decision, because we select  different goals. I want my library to be as close as possible to the  original versions of reflection, so no extra methods and functionality.  This should be a basis for top-level libraries...


There is some misunderstanding here: this is a goal also for BetterReflection, except that we have two separate APIs:

 * one for getting "better" reflection
 * one for emulating core reflection (we reached parity there, btw)

So BetterReflection is still a perfect fit.

lisachenko:
> There is some misunderstanding here: this is a goal also for BetterReflection, except that we have two separate APIs:

>    one for getting "better" reflection
>    one for emulating core reflection (we reached parity there, btw)

> So BetterReflection is still a perfect fit.

Of course, it's our choice as maintainers to depend on specific implementation. Because we should have full control over entire project. When I asked in https://github.com/Roave/BetterReflection/issues/18 about inheritance it wasn't ready and (maybe) team of BR hadn't time for this stuff and I couldn't wait any more. So my own implementation was developed from scratch.

I see a lot of interesting stuff in BR, but in my opinion, it violates SRP for the package and tries to do much more than required for simple reflection, eg. AST transformation, searching for class loaders, etc. BR is kind of Swiss knife, but hey, why I need heavy tool? :) I need fast and simple tool that provides me only Reflection API, it should be enough. All other stuff can be implemented on top of this. This is the reason why I still don't use BR.

There are only few things that missing now in goaop/parser-reflection:
  - to be fully transparent for internal classes.
  - to have an option to specify class name for reflection (kind of factory) to have an ability to build anything more complex on top of this library.

At this point I will finish development and only keep it compatible with new versions of PHP. All other stuff should be composed from existing bricks )

So, from what I can see we will keep running both projects.

Ocramius:
> Of course, it's our choice as maintainers to depend on specific implementation. 

I only rely on the reflection API - the only change here (and you can see it from the diffs) is the initial factory/instantiation of a reflector

malukenho:
@Ocramius can this one be closed?

Ocramius:
This one is still valid from a feature perspective, but should be reduced to removing reflection instantiation where possible.

In practice: patch is not valid anymore, but the feature is still wanted, as it allows for proxying stuff like final classes.

lisachenko:
FYI: Proxying of final classes will be available via [lisachenko/z-engine](https://github.com/lisachenko/z-engine) and PHP>=7.4.

```php
use ZEngine\Reflection\ReflectionClass as ReflectionClassEx;

$refClass = new ReflectionClassEx(SomeFinalClass::class);
$refClass->setFinal(false);
```
After that you can extend your final class easily.

Ocramius:
While that is tempting, it would require adding an extension for something that would otherwise require removing the `final` marker by hand (when you **know** what you are doing).

The initial idea of this patch is to generate proxy code without loading the original sources (which may lead to side-effects)

PR #245, opened 2015-07-20
Removed version requirement in install instruction

pascaldevink:
Composer will do this for you, so there is no reason for it anymore. Also, the * operator is way too wide and composer will probably suggest a better operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ You can learn about the proxy pattern and how to use the **ProxyManager** on the
 The suggested installation method is via [composer](https://getcomposer.org/):
 
 ```sh
-php composer.phar require ocramius/proxy-manager:1.0.*
+php composer.phar require ocramius/proxy-manager
 ```
 
 ## Proxy example


Ocramius:
@pascaldevink does the current logic produce constraints such as `"foo/bar": "1.*"` or `"foo/bar": "1.0.*"`?

pascaldevink:
@Ocramius neither, it will produce ```^1.0``` or ```~1.0``` on older versions

rdohms:
The use of `*` should be limited to situations that run away from semver and require specific versions or ranges to be used. the `~^` operators provide better ways to interact with semver packages, and have a few small performance gains.

Ocramius:
@rdohms I'm not particularly concerned about `*`, what I want is locking at patch-level versions rather than minor versions. If `composer require` doesn't do it, then I will sadly have to reject this patch

rdohms:
@Ocramius indeed the only way to achieve that specific desire is with the star. But it begs the question why? The behaviour you are describing of sticking to patch releases is endemic to Major Version Zero... in which case `^0.1` would work.
But if this is in `1.0` then why are you suggesting the user stick to patch releases and not minor releases as these should be just as safe?

Ocramius:
>  why are you suggesting the user stick to patch releases and not minor releases as these should be just as safe?

Mainly because I may introduce accidental regressions by writing new features. This package is so widely used that such a thing will happen for one of the users somewhere: sticking to a patch release will guarantee that any newly introduced feature doesn't cause changes in the generated code, for example.

rdohms:
@Ocramius yeah then this is indeed a very special case.

But i would say you should have enough tests to be able to do this without such danger.

Ocramius:
@rdohms in order to ensure that, I'd need to run tests from different versions against the current version, and even there, the generated code may have changed (as well as the implemented interfaces and mocks)

Ocramius:
As per discussion, I'm closing the issue as inapplicable. The intent was good, but this package is really better kept safe from eager upgrades.

PR #246, opened 2015-07-24
Don't suggest package that is already required.

WouterSioen:
The zendframework/zend-code dependency already installs zendframework/zend-stdlib. There is thus no need to suggest it as extra library.


Ocramius:
:+1:

PR #248, opened 2015-07-27
Limited requirement of zend-code

bdunogier:
zend-code:2.5.2 bumps the PHP requirement to >= 5.5. Not advisable on a stable version.

Fixes this by limiting the requirement to the last minor that required 5.3.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     ],
     "require": {
         "php":                     ">=5.3.3",
-        "zendframework/zend-code": ">2.2.5,<3.0"
+        "zendframework/zend-code": ">2.2.5,<2.5.2"
     },
     "require-dev": {
         "ext-phar":                    "*",


GrahamCampbell:
:-1:

GrahamCampbell:
Doesn't make sense to prevent people using php 5.5 from using the newer version.

GrahamCampbell:
If you want to force your dependencies to be php 5.3 compatible, say if you want to commit a lock file for a system using an older version of php, then composer has a way to "pretend" to use a an older "platform". It's in their docs somewhere, and composer itself does this so the phar works on php 5.3.

bdunogier:
Hi. Well, there are several things here.

It's not about "forcing" dependencies, it's about changing (or not) the required version of php without warning. Ideally, the fix should be on zend's side, and the php requirement shouldn't change abruptly.

> Doesn't make sense to prevent people using php 5.5 from using the newer version.
That's probably the best argument against a fix in this package, you're right. I'll go annoy zend, then :-)

Cheers, and thank you for the feedback.

GrahamCampbell:
> the fix should be on zend's side, and the php requirement shouldn't change abruptly

It doesn't though, because the version constraint permits using versions that work on php 5.3.

bdunogier:
> It doesn't though, because the version constraint permits using versions that work on php 5.3.

And you're again right. I'll take care of my stuff on my release then ;-)
Thanks again.

GrahamCampbell:
No problem. :)

Issue #249, opened 2015-08-08
Weird problem with FileWriterGeneratorStrategy


twmobius:
I woke up this morning and I found that my apache log was flooded with the following message:

```
[Sat Aug 08 06:30:01.524221 2015] [:error] [pid 4487] [client 127.0.0.1:56176] PHP Warning:  rename(/tmp/ProxyManagerGeneratedProxy__PM__ZendDbResultSetHydratingResultSetGenerated7640b9c8ec8fac4079226d15bfdee2bc.php.55c577b97ff655.54237664,/tmp/ProxyManagerGeneratedProxy__PM__ZendDbResultSetHydratingResultSetGenerated7640b9c8ec8fac4079226d15bfdee2bc.php): Operation not permitted in /var/www/vendor/ocramius/proxy-manager/src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php on line 63
```

What I am guessing might have happened is the following (I am not 100% sure on how the internals of ProxyManager work):

At some point ProxyManager created Proxy ```/tmp/ProxyManagerGeneratedProxy__PM__ZendDbResultSetHydratingResultSetGenerated7640b9c8ec8fac4079226d15bfdee2bc.php``` which for some strange reason had ```root.root``` ownership. (I don't know how this could have happened, apache is running under ```www-data.www-data``` and no restart has taken place at the time this happened. Any ideas welcome)

As a result ```rename``` command on ```FileWriterGeneratorStrategy.php``` could not delete the original file thus creating only a new file and going into some strange loop on the next request.

Thus there were 9000 files on my /tmp folder this morning. I am mentioning it here in case there is something we can do to avoid this from happening in general. (Some check on the rename perhaps or earlier in the proxy generation)

What I don't get is that I've read the tune for production and added the following snippet on ```onBootstrap()```:

```
    $config = new \ProxyManager\Configuration();
    $config->setProxiesTargetDir(__DIR__ . '/../../data/cache/php');
    $config->setAutoGenerateProxies(true);

    spl_autoload_register($config->getProxyAutoloader());
```

But the ProxiesTargetDir is never populated and /tmp has the Proxies instead. Should I manually copy them to the ProxiesTargetDir so that it stops auto-generating them?

Ocramius:
> I woke up this morning and I found that my apache log was flooded

Ouch, sorry about that.

[...]

PR #250, opened 2015-08-08
Hotfix - #249 file writer generator strategy rename failures

Ocramius:
See #249 

This needs to also be rebased and applied to `master` once @twmobius confirms that it solves at least the problem with the temporary files not being removed


malukenho:
:beers:

Ocramius:
@malukenho please also check #251 and #249 when you can

PR #247, opened 2015-07-24
Hotfix: PHP 7 warnings on binding internal classes

Ocramius:

PR #252, opened 2015-08-09
Moving to new travis build infrastructure

Ocramius:
see http://docs.travis-ci.com/user/migrating-from-legacy/


Ocramius:
@malukenho as you can see, `master` is broken regardless of #247

PR #254, opened 2015-08-22
Fix: Methods do not accept as many arguments as passed into them

localheinz:
This PR

* [x] removes a method argument where a method accepts only two arguments, not three
* [x] removes a method argument where a method accepts only one argument, not two
* [x] removes an unused method argument `$proxyOptions`

### DIFF
--- a/src/ProxyManager/Factory/AbstractBaseFactory.php
+++ b/src/ProxyManager/Factory/AbstractBaseFactory.php
@@ -55,11 +55,10 @@ abstract class AbstractBaseFactory
      * Generate a proxy from a class name
      *
      * @param string  $className
-     * @param mixed[] $proxyOptions
      *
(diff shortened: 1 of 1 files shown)


malukenho:
Thanks @localheinz  :beers:

Ocramius:
This should be reverted. Opened #256

Ocramius:
Ref: [`ProxyManager\ProxyGenerator\LazyLoadingGhostGenerator`](https://github.com/Ocramius/ProxyManager/blob/9470713270ea9bcea4a68823400ef619ffafa804/src/ProxyManager/ProxyGenerator/LazyLoadingGhostGenerator.php#L63)

Ocramius:
Reverted at 079399916eb7fba57aed709de2d63f104c696673

PR #261, opened 2015-12-24
Fix a typo in main.css stylesheet

vonalbert:
Removed a double hash '#' character from border color


malukenho:
Thanks @vonalbert :beers:

PR #241, opened 2015-04-09
[WIP] Support scalar type hinting for php7 

malukenho:
Linking #213

PR #271, opened 2016-01-04
Feature - #241 - support scalar type hints

Ocramius:
Relies on zendframework/zend-code#30
Supersedes #241


composer.json
```diff
@@ -19,8 +19,8 @@
         }
     ],
     "require": {
-        "php":                     "~5.6|~7.0",
-        "zendframework/zend-code": "^2.6.1"
+        "php":                     "~7.0",
+        "zendframework/zend-code": "dev-feature/php7-compliance@DEV"
```
Ocramius: To be fixed before merging

PR #277, opened 2016-01-22
Fix/#276 protected member access via friend objects

### LINKED ISSUE #276: Test protected member access lazy loading (ghost objects)
Given following code:

```php
class Foo
{
    private $bar;
    protected $baz;
    public $tab;
    public function getBar(self $other) { return $other->bar; }
    public function getBaz(self $other) { return $other->baz; }
    public function getTab(self $other) { return $other->tab; }
}
```

Any access to these methods, given a non-initialized ghost object, should cause that object to lazy-load.

This should already work, but needs to be tested.

Ocramius:
Fixes #276, provides extensive coverage on the friend objects functionality


tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
```diff
@@ -1070,4 +1073,154 @@ public function skipPropertiesFixture()
...
+        self::assertTrue($proxy->isProxyInitialized());
+        self::assertSame($expectedValue, $accessor($proxy));
+    }
+
+    public function getMethodsThatAccessPropertiesOnOtherObjectsInTheSameScope() : array
+    {
+        $proxyClass = $this->generateProxy(OtherObjectAccessClass::class);
+
+        return [
+            OtherObjectAccessClass::class . '#$privateProperty' => [
```
malukenho: As it's only to describe what you're returning I guess it can be hardcoded

Ocramius: I'm mainly doing this for readability/later refactoring. It gets really complicated to look for `OtherObjectAccessClass` usages otherwise...

Ocramius:
Merging, since @malukenho reviewed :D

Issue #208, opened 2014-12-15
Document new `GhostObject` behavior (property lazy-loading, by-ref initialization)


Ocramius:
See features introduced in #192

Ocramius:
Still needs doing: documentation for this is to be extracted from the test cases, such as:
 * [how the properties are assigned a value](https://github.com/Ocramius/ProxyManager/blob/f7ff39e66f0ab1f62f418eb549994f3c202815fb/tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php#L450-L490)
 * [how the initialization process actually works](https://github.com/Ocramius/ProxyManager/blob/f7ff39e66f0ab1f62f418eb549994f3c202815fb/tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php#L629-L655)
 * [the fact that a static constructor is added to the proxy](https://github.com/Ocramius/ProxyManager/blob/f7ff39e66f0ab1f62f418eb549994f3c202815fb/tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php#L320)
 * [how method calls that don't access internal state will not initialize a proxy](https://github.com/Ocramius/ProxyManager/blob/f7ff39e66f0ab1f62f418eb549994f3c202815fb/tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php#L85-L120)

Ocramius:
Handled in #281

PR #281, opened 2016-01-22
Docs - #208 document new ghost object behavior

Ocramius:
Provides docs for #208


fntlnz:
Form is OK,  the only thing that's unclear to me is that if in the end there's any way to not skip the constructor.

Ocramius:
@fntlnz you can actually call the constructor, if you want, but since we're setting all the state, is that an actual use-case?

Ocramius:
@fntlnz added block to clarify that :-)

fntlnz:
@Ocramius at the moment I don't see any use case for that. But probably because is stated "the constructor is not called" one may think  something like: "hey I have to proxy that old legacy painful object with a lot of business logic poured into the controller and I want to actually call it to be sure that the things are going as before but here's it's written that the controller isn't called. Unfortunately I can't use this tool :( ". That's quite a edge case and son of poorly designed code of course but, shit happens.
 Thanks for clarifying that :)

Ocramius:
@fntlnz thanks for bringing it up :-)

PR #291, opened 2016-01-31
Test - cloned properties should be preserved

Ocramius:
Just adds a check for cloned properties: should preserve them.


Ocramius:
Note: should be merged also into 2.0.x, and 2.0.1 should be released afterwards.

Issue #298, opened 2016-03-02
Setting value on proxy via ReflectionProperty causes an error.


bendavies:
Hi there,

I have a lazy service.

ProxyManger generates a `__set` method (via `ProxyManager\ProxyGenerator\UtilPublicScopeSimulator`) that looks like:
```
    public function __set($name, $value)
    {
        $this->initializer56d6e5ca17346071728280 && $this->initializer56d6e5ca17346071728280->__invoke($this->valueHolder56d6e5ca17340889958830, $this, '__set', array('name' => $name, 'value' => $value), $this->initializer56d6e5ca17346071728280);
        $realInstanceReflection = new \ReflectionClass(get_parent_class($this));
        if (! $realInstanceReflection->hasProperty($name)) {
            $targetObject = $this->valueHolder56d6e5ca17340889958830;
            return $targetObject->$name = $value;
            return;
        }
        $targetObject = $this->valueHolder56d6e5ca17340889958830;
        $accessor = function & () use ($targetObject, $name, $value) {
            return $targetObject->$name = $value;
        };
            $backtrace = debug_backtrace(true);
            $scopeObject = isset($backtrace[1]['object']) ? $backtrace[1]['object'] : new \ProxyManager\Stub\EmptyClassStub();
            $accessor = $accessor->bindTo($scopeObject, get_class($scopeObject));
        $returnValue = & $accessor();
        return $returnValue;
    }
```

If you then set a value via `ReflectionProperty` on this proxy, you will receive an error:
> Cannot bind closure to scope of internal class ReflectionProperty

on the `bindTo` call.

Let me know if you need anything else?

Ocramius:
Argh, php limitations.

Could you write an example, following the ones in https://github.com/Ocramius/ProxyManager/tree/master/tests/language-feature-scripts ?

bendavies:
yep, sure. will do

bendavies:
here you go
```
--TEST--
Verifies that generated access interceptors allow protected property write via Reflection
--FILE--
<?php

require_once __DIR__ . '/init.php';

class Kitchen
{
    protected $sweets;
}

$factory = new \ProxyManager\Factory\AccessInterceptorValueHolderFactory($configuration);

$proxy = $factory->createProxy(new Kitchen());

$reflO = new \ReflectionObject($proxy);
$reflP = $reflO->getProperty('sweets');

$reflP->setAccessible(true);
$reflP->setValue($proxy, 'stolen');

[...]

PR #299, opened 2016-03-07
testcase for #298 - limitation - lazy loading value holder protected properties can be accessed via reflection

bendavies:
ref #298

### DIFF
--- a/tests/language-feature-scripts/lazy-loading-value-holder-allows-protected-property-reflection-write.phpt
+++ b/tests/language-feature-scripts/lazy-loading-value-holder-allows-protected-property-reflection-write.phpt
@@ -1,0 +1,26 @@
+--TEST--
+Verifies that generated access interceptors allow protected property write via Reflection
+--FILE--
+<?php
+
+require_once __DIR__ . '/init.php';
+
+class Kitchen
+{
+    protected $sweets;
+}
+
+$factory = new \ProxyManager\Factory\AccessInterceptorValueHolderFactory($configuration);
+
+$proxy = $factory->createProxy(new Kitchen());
+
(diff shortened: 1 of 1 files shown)


tests/language-feature-scripts/lazy-loading-ghost-allows-protected-property-reflection-write.phpt
```diff
@@ -0,0 +1,26 @@
...
+<?php
+
+require_once __DIR__ . '/init.php';
+
+class Kitchen
+{
+    protected $sweets;
+}
+
+$factory = new \ProxyManager\Factory\AccessInterceptorValueHolderFactory($configuration);
```
Ocramius: Just noticed that the file name does not match the test case... Indeed, this is a value holder issue, not a ghost object isse

bendavies: renamed :)

Ocramius: Thanks!

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

Ocramius:
Looked into this tonight: this would basically require emulation of what is going on in the lazy-loading ghost objects, which unset all properties, regardless if public or private.

That's a massive rewrite of the functionality, so it may only happen in 3.x

bendavies:
@Ocramius just wondering if you are planning this for 3.x?

Ocramius:
@bendavies I didn't plan it. 3.x may be just some 7.1 support, for now.

Issue #300, opened 2016-03-13
Null objects do not extend their original class


trickleup:
It seems to me that classes with PHP7 return types fail.

```
PHP Fatal error:  Declaration of ProxyManagerGeneratedProxy\__PM__\MyClass::queryUsers() must be compatible with MyClass::queryUsers(): array in /Users/johan/Projects/sparkcentral/auth-context/vendor/ocramius/proxy-manager/src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php(63) : eval()'d code on line 3
```

Is there a workaround/fix for this issue?

trickleup:
Also, the same seems to be the case with scalar type hints in method declarations.

Ocramius:
@trickleup ProxyManager v2 fully supports return type hints: can you check what version you have installed?

michaelmoussa:
@Ocramius I'm having (_what I think is_) the same issue, only with the `Null Object Proxy`. See example case below:

```
$ cat test.php
<?php

declare (strict_types = 1);

require __DIR__ . '/vendor/autoload.php';

echo \ProxyManager\Version::getVersion() . "\n";

class Foo
{
}

function bar() : Foo
{
    return (new \ProxyManager\Factory\NullObjectFactory())->createProxy(Foo::class);
}

bar();
```

```
$ php test.php
2.0.1@6c89b7bd6039d8047b1473e2074cb56baa4bc15d
PHP Fatal error:  Uncaught TypeError: Return value of bar() must be an instance of Foo, instance of ProxyManagerGeneratedProxy\__PM__\Foo\Generatede07f73e523c2fb4772e096fa7615d794 returned in /Users/mmoussa/Code/pm-test/test.php:15
```

The generated code looks like this:

```php
namespace ProxyManagerGeneratedProxy\__PM__\Foo;

class Generatede07f73e523c2fb4772e096fa7615d794 implements \ProxyManager\Proxy\NullObjectInterface
{

[...]

PR #301, opened 2016-03-16
Correcting null object inheritance - null objects were not extending from their original class

michaelmoussa:
@Ocramius first stab at addressing #300 for Null Objects.

### DIFF
--- a/src/ProxyManager/ProxyGenerator/NullObjectGenerator.php
+++ b/src/ProxyManager/ProxyGenerator/NullObjectGenerator.php
@@ -51,11 +51,13 @@ class NullObjectGenerator implements ProxyGeneratorInterface
 
         if ($originalClass->isInterface()) {
             $interfaces[] = $originalClass->getName();
+        } else {
+            $classGenerator->setExtendedClass($originalClass->getName());
         }
 
         $classGenerator->setImplementedInterfaces($interfaces);
 
-        foreach (ProxiedMethodsFilter::getProxiedMethods($originalClass) as $method) {
+        foreach (ProxiedMethodsFilter::getProxiedMethods($originalClass, []) as $method) {
             $classGenerator->addMethodFromGenerator(
                 NullObjectMethodInterceptor::generateMethod(
                     new MethodReflection($method->getDeclaringClass()->getName(), $method->getName())
--- a/tests/ProxyManagerTest/ProxyGenerator/NullObjectGeneratorTest.php
+++ b/tests/ProxyManagerTest/ProxyGenerator/NullObjectGeneratorTest.php
@@ -80,7 +80,7 @@ class NullObjectGeneratorTest extends AbstractProxyGeneratorTest
         $proxyGenerated = $generatedClassName::staticProxyConstructor();
 
         foreach (Properties::fromReflectionClass($generatedReflection)->getPublicProperties() as $property) {
-            $this->assertNull($proxyGenerated->$property);
+            $this->assertNull($proxyGenerated->{$property->getName()});
         }
 
         foreach ($generatedReflection->getMethods(ReflectionMethod::IS_PUBLIC) as $method) {
--- a/tests/language-feature-scripts/null-object-satisfies-return-type-declaration.phpt
+++ b/tests/language-feature-scripts/null-object-satisfies-return-type-declaration.phpt
@@ -1,0 +1,25 @@
+--TEST--
+Verifies that generated null object satisfies return type declarations
+--FILE--
+<?php
+
+declare (strict_types = 1);
+
+require_once __DIR__ . '/init.php';
+
+class Kitchen
+{
+    public function foo()
+    {
+        return 'bar';
+    }
+}
+
+$factory = new \ProxyManager\Factory\NullObjectFactory($configuration);
+
+$proxy = $factory->createProxy(Kitchen::class);
+
+var_dump($proxy instanceof Kitchen);
+?>
+--EXPECT--
+bool(true)


src/ProxyManager/ProxyGenerator/NullObjectGenerator.php
```diff
@@ -51,11 +51,13 @@ public function generate(ReflectionClass $originalClass, ClassGenerator $classGe
...
         if ($originalClass->isInterface()) {
             $interfaces[] = $originalClass->getName();
+        } else {
+            $classGenerator->setExtendedClass($originalClass->getName());
         }
 
         $classGenerator->setImplementedInterfaces($interfaces);
 
-        foreach (ProxiedMethodsFilter::getProxiedMethods($originalClass) as $method) {
+        foreach (ProxiedMethodsFilter::getProxiedMethods($originalClass, []) as $method) {
```
Ocramius: Why was this change necessary?

tests/ProxyManagerTest/ProxyGenerator/NullObjectGeneratorTest.php
```diff
@@ -80,7 +80,7 @@ public function testGeneratesValidCode(string $className)
         $proxyGenerated = $generatedClassName::staticProxyConstructor();
 
         foreach (Properties::fromReflectionClass($generatedReflection)->getPublicProperties() as $property) {
-            $this->assertNull($proxyGenerated->$property);
+            $this->assertNull($proxyGenerated->{$property->getName()});
```
Ocramius: Urgh, was this code not run by the tests? :|

src/ProxyManager/ProxyGenerator/NullObjectGenerator.php
michaelmoussa: Removing it causes this failure:

```
1) ProxyManagerTest\ProxyGenerator\NullObjectGeneratorTest::testGeneratesValidCode with data set #2 ('ProxyManagerTestAsset\ClassWi...ethods')
Only variable references should be returned by reference

/Users/mmoussa/Code/ocramius/ProxyManager/tests/ProxyManagerTestAsset/ClassWithByRefMagicMethods.php:68
/Users/mmoussa/Code/ocramius/ProxyManager/tests/ProxyManagerTest/ProxyGenerator/NullObjectGeneratorTest.php:88
```

The line in question is:

```php
$this->assertNull(call_user_func([$proxyGenerated, $method->getName()]));
```

`ClassWithByRefMagicMethods` is the dataset that causes it to fail, specifically the `__sleep` method.

`ProxiedMethodsFilter::getProxiedMethods(...)` will exclude the following methods if no `$excluded` param is provided:

```
        '__get',
        '__set',
        '__isset',
        '__unset',
        '__clone',
        '__sleep',
        '__wakeup',
```

Changing adding the empty array parameter causes `__sleep` (and `__wakeup`, which would have caused the failure if `__sleep` hadn't), fixes it.

I haven't used ProxyManager too extensively so I wasn't sure of the overall impact of this change, _but_ it didn't seem to break any existing tests, so I figured we could evaluate further during review. I suppose it _could_ be considered a BC break, though, if someone was relying on the default `null` value to not proxy the magic methods, _and_ they were returning references.

However, after looking at this more closely, maybe it's not as straightforward...

The two methods that would cause the error in the `ClassWithByRefMagicMethods` class are:

```php
    public function & __sleep()
    {
    }

    public function & __wakeup()
    {
    }
```

And neither of those methods in that class returns a value... but their not returning a value _isn't_ the issue, this is:

```
        foreach ($generatedReflection->getMethods(ReflectionMethod::IS_PUBLIC) as $method) {
            if (! ($method->getNumberOfParameters() || $method->isStatic())) {
                $this->assertNull(call_user_func([$proxyGenerated, $method->getName()]));
            }
        }
```

The other methods in that class don't get called at all for this test because they have parameters. `__sleep` and `__wakeup` do get called because they have no params, and since they don't return a value, the test encounters an error. This wasn't an issue before my PR because, before my PR, the generated null objects weren't extending the original class (the fixing of which was the entire point of the PR). Since they didn't extend the original class, they wouldn't be run as part of that reflection method loop in the test class.

Now, since they _do_ extend the original class (and the magic methods are not proxied by default), the `__sleep` and `__wakeup` are being called.

So, the two options that come to mind are:

1. Leave my `[]` argument addition change. Since we're creating a proxy, we presumably want to proxy all methods, right?
2. Do not run the `$this->assertNull(call_user_func([$proxyGenerated, $method->getName()]));` assertion and surrounding loop for the `ClassWithByRefMagicMethods` dataset. Before my PR, those methods weren't being executed anyway.

The second option might be better, since that one doesn't change the library code beyond the original point of the PR (to fix typehints / return types in PHP 7 for the generated proxy null objects), and it doesn't change the _actual_ behavior of the test, because that portion of the test didn't get executed at all in the original version.

tests/ProxyManagerTest/ProxyGenerator/NullObjectGeneratorTest.php
michaelmoussa: Similar issue to what I mentioned in my other comment. :) To illustrate, check out `master`, leave only `[BaseClass::class]` and `[ClassWithMixedProperties::class]` in the dataprovider, and add a `die` right before `$this->assertNull($proxyGenerated->$property);`.

That assertion only gets run if it's able to reflect any properties, which it doesn't because the original class isn't extended.

Ocramius:
@michaelmoussa 👍 \o/

PR #313, opened 2016-05-15
Documentation fixlets

machour:
As promised!

PR #315, opened 2016-05-24
Update lazy-loading-value-holder.md

holtkamp:


Ocramius:
This change was already applied via #313. Thanks anyway! \o

PR #327, opened 2016-07-24
PHP 7.1 support

Ocramius:
Note: marked this as 3.0.0 because I plan to drop 7.0 support as soon as this is in. The reasoning behind that is mainly reducing version support to simplify maintenance: no real BC breaks here.

TODO:
- [x] patch https://github.com/zendframework/zend-code/blob/e21f53c1f8a40dd41245eed1f9887f61a675517b/src/Generator/TypeGenerator.php to support `void` (should be a one-liner) ( https://github.com/zendframework/zend-code/issues/84 )
- [x] test newly introduced conditionals
- [x] support nullable parameter types (requires patching zend-code again) ( https://github.com/zendframework/zend-code/issues/85 )
- [x] support nullable return types (requires patching zend-code again) ( https://github.com/zendframework/zend-code/issues/85 )
- [x] refactor `void` return type checker into a static utility
- [x] refactor "return statement generator" away
- [x] wait for [`zendframework/zend-code:3.1.0`](https://github.com/zendframework/zend-code/milestone/11)
- [x] functional tests with `void` for following code generators:
  - [x] ghost object ocramius/ProxyManager@4ae5727936a46acd27e1a0db77181b2844e7a2e8
  - [x] lazy loading value holder ocramius/ProxyManager@503266c841945976160fd2bb527ac8138c620d94
  - [x] access interceptor value holder ocramius/ProxyManager@ff3d5be447ca886c408828abbccf3dbe158872ee
  - [x] access interceptor scope localizer ocramius/ProxyManager@3bbbdd8639133a07f540b10c7ac4e246e0cb142f
  - [x] remote object ocramius/ProxyManager@e58efeb879dfb8173757375c5a685d8a0b98b8cc
  - [x] null object ocramius/ProxyManager@0f00756a2301f0689124ad07dd67f9cead5abba0
- [x] ~~write a generic type checker that compares reflection signatures of parent/child classes (to be run against ALL code generators)~~ out of scope


src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/Util/InterceptorGenerator.php
```diff
@@ -75,9 +76,20 @@ public static function createInterceptedMethodBody(
             . "    \$suffixReturnValue = \$this->$suffixInterceptors" . "[$name]->__invoke("
             . "\$this, \$this, $name, $paramsString, \$returnValue, \$returnEarly);\n\n"
             . "    if (\$returnEarly) {\n"
-            . "        return \$suffixReturnValue;\n"
+            . '        ' . self::returnStatement('$suffixReturnValue', $originalMethod)
             . "    }\n"
             . "}\n\n"
-            . "return \$returnValue;";
+            . self::returnStatement('$returnValue', $originalMethod);
```
malukenho: `rtrim(self::returnStatement('$returnValue', $originalMethod))` on this last statement

src/ProxyManager/ProxyGenerator/AccessInterceptorValueHolder/MethodGenerator/Util/InterceptorGenerator.php
```diff
@@ -78,9 +79,20 @@ public static function createInterceptedMethodBody(
             . "    \$suffixReturnValue = \$this->$suffixInterceptors" . "[$name]->__invoke("
             . "\$this, \$this->$valueHolder, $name, $paramsString, \$returnValue, \$returnEarly);\n\n"
             . "    if (\$returnEarly) {\n"
-            . "        return \$suffixReturnValue;\n"
+            . '        ' . self::returnStatement('$suffixReturnValue', $originalMethod)
             . "    }\n"
             . "}\n\n"
-            . "return \$returnValue;";
+            . self::returnStatement('$returnValue', $originalMethod);
```
malukenho: `rtrim(self::returnStatement('$returnValue', $originalMethod))` here too

stof:
@Ocramius dropping support for PHP 7.0 in the same release introducing support for PHP 7.1 makes it much harder for projects having to support PHP 7.0 and 7.1, as they cannot select a version working for both.

Ocramius:
@stof I won't change any API for `3.0.0`, from what I can see. Allowing `^2.0.0 || ^3.0.0` will be sufficient.

I just don't want to maintain a codebase that does both codegen for 7.0 and 7.1, since the number of tiny annoying details leads to a massive amount of pain.

Ocramius:
Last standing issue is https://github.com/phpbench/phpbench/issues/450 - other than that, we can merge this!

Ocramius:
@stof I re-targeted this to `2.1.0`, since no API changes are included, but PHP 7.0 support was dropped.

PR #329, opened 2016-08-01
Fix incorrect type in docs example

GaryJones:
Docs example says closure arg `$parameters` is a string, when I think it should be an array.


GaryJones:
Pretty sure the build error is not due to my diff.

Ocramius:
@GaryJones indeed, failure unrelated, will inspect.

PR #331, opened 2016-08-20
Turned closure into local variable due to some serialization issues.

shochdoerfer:
After upgrading [Disco](https://github.com/bitexpert/disco) to Proxy Manager 2, I ran into an issue with the FileWriterGeneratorStrategy class. When trying to serialize the Beanfactory Configuration class which can potentially hold a FileWriterGeneratorStrategy instance PHP will throw an error because one of the properties of FileWriterGeneratorStrategy is a closure which cannot be serialized. 

Thus I turned the closure into a local variable to work around the issue. I could not see any side-effects, but maybe there was a reason for this in the first place.

### DIFF
--- a/src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
+++ b/src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
@@ -40,18 +40,11 @@ class FileWriterGeneratorStrategy implements GeneratorStrategyInterface
     protected $fileLocator;
 
     /**
-     * @var callable
-     */
-    private $emptyErrorHandler;
-
-    /**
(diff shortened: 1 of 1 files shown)


Ocramius:
@shochdoerfer I understand why you want this, but this is a service, and should never be serialized anyway (I should add `private function __sleep() {}` to all of them), and I cannot guarantee for it to be functioning after serialisation.

Can you point me to a test that fails because of this, please?

shochdoerfer:
@Ocramius sure will do.

shochdoerfer:
@Ocramius I could fix the issue in Disco itself. Had previously stored the Proxy Manager config internally and thus run into the serialization issue. Found a way to get rid of it.

Ocramius:
@shochdoerfer thanks for fixing it on your end!

PR #341, opened 2016-10-09
Removed the reference to `$emptyErrorHandler`

### LINKED ISSUE #340: Unable to serialize proxy factory, when using FileWriterGeneratorStrategy as cache
At the moment, the `FileWriterGeneratorStrategy` sets and resets the error handler to an empty closure. The problem arises, when trying to serialize a proxy factory, which uses it as a caching mechanism. 

The serialization fails with `Uncaught Exception: Serialization of 'Closure' is not allowed`

DaGhostman:
Closes #340 since the object no longer holds a reference to the empty closure

### DIFF
--- a/src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
+++ b/src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
@@ -40,18 +40,11 @@ class FileWriterGeneratorStrategy implements GeneratorStrategyInterface
     protected $fileLocator;
 
     /**
-     * @var callable
-     */
-    private $emptyErrorHandler;
-
-    /**
      * @param \ProxyManager\FileLocator\FileLocatorInterface $fileLocator
      */
     public function __construct(FileLocatorInterface $fileLocator)
     {
         $this->fileLocator       = $fileLocator;
-        $this->emptyErrorHandler = function () {
-        };
     }
 
     /**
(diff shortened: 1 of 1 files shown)


DaGhostman:
I think the last error of the build is not related to the changes made.

Ocramius:
The reason for the existence of `$emptyErrorHandler` is to avoid useless GC and closure instantiation.

Why do you need to serialize this class? It's not supposed to be serialized, and if this sort of PR keeps popping up, I'll start filling the entire library with `public function __sleep() { throw new DontDoThis(); }`

DaGhostman:
Well at the moment I am working on a small DI container, which wraps the produced objects (if necessary) in a proxy to provide AOP like functionality before and After method calls an dproperty access.

When I pass the proxy to the container I need to persist it, since when the container gets unserialized (caching) it still needs the proxy factory in order to wrap further objects.. 

Any suggestions on a different approach are welcome?

Ocramius:
> Any suggestions on a different approach are welcome?

@shochdoerfer you had a similar issue, can you eventually point @DaGhostman to a solution?

@DaGhostman in general, you should not serialize components of this library. I opened #342 to tackle this issue with better error messages, for future reference.

Closing as `invalid`

shochdoerfer:
@DaGhostman Had the same issue in [bitexpert/disco](https://github.com/bitexpert/disco) which in fact is also a DI container ;) In the previous version I tried to serialize the whole DI container and ran into similar issues like the one you are pointing out. The solution was to create a separate "store" for the objects and only allow the serialization of the store, see [here](https://github.com/bitexpert/disco#session-aware-beans) for a high-level description.

PR #348, opened 2016-11-20
Introduced full mutation testing

Ocramius:
3e7fcd


Ocramius:
"You need to install and enable xdebug in order to allow for code coverage generation."

Dang.

src/ProxyManager/Exception/FileNotWritableException.php
```diff
@@ -39,24 +39,4 @@ public static function fromInvalidMoveOperation(string $fromPath, string $toPath
             $toPath
         ));
     }
-
-    public static function fromNonWritableLocation($path) : self
```
Ocramius: Actually, as a better alternative, I should deprecate this method, then restore the tests and drop it in 3.0.

samsonasik:
@Ocramius I created PR to your PR at #349 to install xdebug manually at php 7.1 env

Ocramius:
Waiting for @travis-ci, then merging and possibly releasing.

Ocramius:
\o/

PR #350, opened 2016-11-22
[WIP] Fix #346 - magic methods hints need to be reflected in generated child classes

### LINKED ISSUE #346: Generated proxies don't respect magic methods' inherited signatures
ProxyManager will raise multiple notices due to class signature incompatibility when proxying a class like following one:

```php
class ClassWithTypedMagicMethods
{
    public function __get(string $name) : string { return 'foo'; }
    public function __set(string $name, string $value) : string { return 'foo'; }
    public function __unset(string $name) : bool { return true; }
    public function __isset(string $name) : bool { return true; }
}
```

A class generation/evaluation test is sufficient for this scenario. The issue is simple: class signature should be copied, when available, instead of being constructed from scratch.

Ocramius:
This is just a patch that gets us where we want, but the introduced complexity is too much, therefore will need to find a better way to implement it, without having codegen talking to every reflection layer.

### DIFF
--- a/src/ProxyManager/Generator/MagicMethodGenerator.php
+++ b/src/ProxyManager/Generator/MagicMethodGenerator.php
@@ -21,6 +21,8 @@ declare(strict_types=1);
 namespace ProxyManager\Generator;
 
 use ReflectionClass;
+use Zend\Code\Generator\ParameterGenerator;
+use Zend\Code\Reflection\ParameterReflection;
 
 /**
  * Method generator for magic methods
@@ -31,24 +33,55 @@ use ReflectionClass;
 class MagicMethodGenerator extends MethodGenerator
 {
     /**
-     * @param ReflectionClass $originalClass
-     * @param string          $name
-     * @param array           $parameters
+     * @param ReflectionClass      $originalClass
+     * @param string               $name
+     * @param ParameterGenerator[] $parameters
+     *
+     * @throws \Zend\Code\Generator\Exception\InvalidArgumentException
      */
     public function __construct(ReflectionClass $originalClass, string $name, array $parameters = [])
     {
-        parent::__construct(
-            $name,
-            $parameters,
-            static::FLAG_PUBLIC,
-            null,
-            $originalClass->hasMethod($name) ? '{@inheritDoc}' : null
-        );
+        parent::__construct($name, [], static::FLAG_PUBLIC);
 
         $this->setReturnsReference(strtolower($name) === '__get');
 
         if ($originalClass->hasMethod($name)) {
-            $this->setReturnsReference($originalClass->getMethod($name)->returnsReference());
+            $originalMethod = $originalClass->getMethod($name);
+
+            $this->setDocBlock('{@inheritDoc}');
+            $this->setReturnsReference($originalMethod->returnsReference());
+            $this->setReturnType(((string) $originalMethod->getReturnType()) ?: null);
+            $this->addOrReplaceParameterNames($originalMethod, ...$parameters);
+        } else {
+            $this->setParameters($parameters);
         }
+    }
+
+    private function addOrReplaceParameterNames(
+        \ReflectionMethod $originalMethod,
+        ParameterGenerator ...$newParameters
+    ) : void {
+        $originalParameters  = array_values(array_map(
+            function (\ReflectionParameter $parameter) use ($originalMethod) : ParameterGenerator {
+                return ParameterGenerator::fromReflection(new ParameterReflection(
+                    [$originalMethod->getDeclaringClass()->getName(), $originalMethod->getName()],
+                    $parameter->getPosition()
+                ));
+            },
+            $originalMethod->getParameters()
+        ));
+
+        $parametersCount = max(count($originalParameters), count($newParameters));
+        $parameters      = [];
+
+        for ($idx = 0; $idx < $parametersCount; $idx += 1) {
+            $parameter = $originalParameters[$idx] ?? $newParameters[$idx];
+
+            $parameter->setName(($newParameters[$idx] ?? $originalParameters[$idx])->getName());
+
+            $parameters[] = $parameter;
+        }
+
+        $this->setParameters($parameters);
     }
 }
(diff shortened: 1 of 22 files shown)


src/ProxyManager/Generator/MagicMethodGenerator.php
```diff
@@ -31,24 +33,55 @@
...
 
         if ($originalClass->hasMethod($name)) {
-            $this->setReturnsReference($originalClass->getMethod($name)->returnsReference());
+            $originalMethod = $originalClass->getMethod($name);
+
+            $this->setDocBlock('{@inheritDoc}');
+            $this->setReturnsReference($originalMethod->returnsReference());
+            $this->setReturnType(((string) $originalMethod->getReturnType()) ?: null);
+            $this->addOrReplaceParameterNames($originalMethod, ...$parameters);
+        } else {
```
Ocramius: Don't like this `else`. Can probably just call the parent constructor with the given parameters, then replace

src/ProxyManager/Generator/MagicMethodGenerator.php
```diff
@@ -31,24 +33,55 @@
...
+            $this->setParameters($parameters);
         }
     }
+
+    private function addOrReplaceParameterNames(
```
Ocramius: This method is way too complex: need to split into multiple private methods explaining exactly what is going on at each step.

src/ProxyManager/Generator/MagicMethodGenerator.php
```diff
@@ -31,24 +33,55 @@
...
+        \ReflectionMethod $originalMethod,
+        ParameterGenerator ...$newParameters
+    ) : void {
+        $originalParameters  = array_values(array_map(
+            function (\ReflectionParameter $parameter) use ($originalMethod) : ParameterGenerator {
+                return ParameterGenerator::fromReflection(new ParameterReflection(
```
Ocramius: Loads of coupling due to Zend\Code's shitty reflection API :-(

src/ProxyManager/ProxyGenerator/LazyLoadingGhost/MethodGenerator/MagicUnset.php
```diff
@@ -126,30 +127,40 @@ public function __construct(
...
-            $publicProperties->getName(),
-            $protectedProperties->getName(),
-            $protectedProperties->getName(),
-            $privateProperties->getName(),
-            $privateProperties->getName(),
-            $privateProperties->getName(),
-            $parentAccess
+
+        $symbols = [
+            '%init'                 => '$this->' . $initializerProperty->getName()
```
Ocramius: I should hunt down all the `sprintf()` and replace them with this syntax: much easier to follow.

src/ProxyManager/ProxyGenerator/LazyLoadingValueHolder/MethodGenerator/MagicGet.php
```diff
@@ -54,7 +54,7 @@ public function __construct(
     ) {
         parent::__construct($originalClass, '__get', [new ParameterGenerator('name')]);
 
-        $hasParent = $originalClass->hasMethod('__get');
+        $hasParent = $originalClass->hasMethod('__get') ? $originalClass->getMethod('__get') : null;
```
Ocramius: This decision path is duplicated everywhere, and can surely be simplified.

src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
```diff
@@ -42,19 +43,21 @@ class PublicScopeSimulator
...
+     *                                                   to read the property. `$this` if none provided
+     * @param string|null            $returnPropertyName name of the property to which we want to assign the result of
+     *                                                   the operation. Return directly if none provided
      *
      * @return string
      *
      * @throws \InvalidArgumentException
      */
     public static function getPublicAccessSimulationCode(
+        ?\ReflectionMethod $originalMethod,
```
Ocramius: BC break.

Ocramius: Also note: since we are passing down reflection details, we may as well get rid of a lot of runtime boilerplate, to be replaced with string constants.

src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
```diff
@@ -147,30 +149,32 @@ private static function getTargetObject(PropertyGenerator $valueHolder = null) :
...
+                );
             case static::OPERATION_ISSET:
-                return 'return isset($targetObject->$' . $nameParameter . ');';
+                return ProxiedMethodReturnExpression::generate(
+                    'isset($targetObject->$' . $nameParameter . ')',
+                    $originalMethod
+                );
             case static::OPERATION_UNSET:
-                return 'unset($targetObject->$' . $nameParameter . ');';
+                return 'unset($targetObject->$' . $nameParameter . ');'
```
Ocramius: `return unset($foo)` is not a valid expression, so I simply add a fake return block afterwards.

tests/ProxyManagerTest/Functional/MultipleProxyGenerationTest.php
```diff
@@ -119,6 +121,8 @@ public function getTestedClasses() : array
         return [
             [BaseClass::class],
             [ClassWithMagicMethods::class],
+            [ClassWithHintedMagicMethods::class],
+            //[InterfaceWithHintedMagicMethods::class],
```
Ocramius: This test doesn't verify interface proxying - need to move to appropriate functional tests instead.

Majkl578:
2 years, any progress?

Two more assets that make test suite crash: https://github.com/Majkl578/ProxyManager/commit/06d9f51c1ee7191ee577f270ee41e8f24ba960de

Note that with 7.2+ requirement you could simply ignore parameters, but you still need return types and void. :trollface:

Ocramius:
No progress, but I could restart it. can you PR the assets to `master`?

Majkl578:
#427

PR #357, opened 2016-12-21
fix notice on __get error when no trace function/file is available (internal PHP functions/methods)

sakarikl:
If backtrace does not have file or line data error string generating caused notice

### DIFF
--- a/src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
+++ b/src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
@@ -108,8 +108,8 @@ class PublicScopeSimulator
             . '            \'Undefined property: %s::$%s in %s on line %s\',' . "\n"
             . '            get_parent_class($this),' . "\n"
             . '            $' . $nameParameter . ',' . "\n"
-            . '            $backtrace[0][\'file\'],' . "\n"
-            . '            $backtrace[0][\'line\']' . "\n"
+            . '            $backtrace[0][\'file\'] ?? \'\',' . "\n"
+            . '            $backtrace[0][\'line\'] ?? \'\'' . "\n"
             . '        ),' . "\n"
             . '        \E_USER_NOTICE' . "\n"
             . '    );' . "\n";
--- a/tests/ProxyManagerTest/ProxyGenerator/Util/PublicScopeSimulatorTest.php
+++ b/tests/ProxyManagerTest/ProxyGenerator/Util/PublicScopeSimulatorTest.php
@@ -50,8 +50,8 @@ if (! $realInstanceReflection->hasProperty($foo)) {
             'Undefined property: %s::$%s in %s on line %s',
             get_parent_class($this),
             $foo,
-            $backtrace[0]['file'],
-            $backtrace[0]['line']
+            $backtrace[0]['file'] ?? '',
+            $backtrace[0]['line'] ?? ''
         ),
         \E_USER_NOTICE
     );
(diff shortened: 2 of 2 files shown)


src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
```diff
@@ -108,8 +108,8 @@ private static function getUndefinedPropertyNotice(string $operationType, string
             . '            \'Undefined property: %s::$%s in %s on line %s\',' . "\n"
             . '            get_parent_class($this),' . "\n"
             . '            $' . $nameParameter . ',' . "\n"
-            . '            $backtrace[0][\'file\'],' . "\n"
-            . '            $backtrace[0][\'line\']' . "\n"
+            . '            $backtrace[0][\'file\'] ?? \'\',' . "\n"
```
Ocramius: In which scenarios would this be the case? Can you add a functional test (`.phpt`)?

Ocramius (changes requested):

src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
sakarikl: Code like this:

`
class Foo {
    private $value = 'Value';
}

class Bar  extends Foo{
}

$factory = new \ProxyManager\Factory\LazyLoadingValueHolderFactory();
$proxy = $factory->createProxy(
    \Bar::class,
    function (& $wrappedObject, $proxy, $method, $parameters, & $initializer) {
        $wrappedObject = new Bar(); // instantiation logic here
        $initializer   = null; // turning off further lazy initialization
    }
);

$r = new \ReflectionClass(\Foo::class);
$p = $r->getProperty('value');
$p->setAccessible(true);

$p->getValue($proxy);
`

causes output:

PHP Notice:  Undefined index: file in /vendor/ocramius/proxy-manager/src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php(70) : eval()'d code on line 85

PHP Notice:  Undefined index: line in /vendor/ocramius/proxy-manager/src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php(70) : eval()'d code on line 86

PHP Notice:  Undefined property: Bar::$value in  on line  in /vendor/ocramius/proxy-manager/src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php(70) : eval()'d code on line 88

Ocramius:
Ah yes, internal reflection produces no file/line in a trace, good point!

Ocramius:
While trying to check for a suitable test case, I found that this actually hits a known edge case ( https://github.com/Ocramius/ProxyManager/pull/299 ). Currently unsure if fixable, since the code should work without failures in this scenario:

```php
class Foo {
    private $value = 'Value';
}

class Bar  extends Foo{
}

$factory = new \ProxyManager\Factory\LazyLoadingValueHolderFactory();
$proxy = $factory->createProxy(
    \Bar::class,
    function (& $wrappedObject, $proxy, $method, $parameters, & $initializer) {
        $wrappedObject = new Bar(); // instantiation logic here
        $initializer   = null; // turning off further lazy initialization
    }
);

$r = new \ReflectionClass(\Foo::class);
$p = $r->getProperty('value');
$p->setAccessible(true);

$p->getValue($proxy);
```

Ocramius:
Indeed, I cannot reproduce this outside of this scenario (which is already broken).

Tried various proxy types, but this one seems to be the only affected one...

PR #361, opened 2017-01-09
PHPStan inspection fixes

Ocramius:


ondrejmirtes:
BTW if you encounter an error about `parent` in return typehint, [I fixed that in dev-master](https://github.com/phpstan/phpstan/commit/764864961a2415e5f9f8790d4624bdf0109303fc). 

Also, [I added a check for `parent` typehint](https://github.com/phpstan/phpstan/commit/763ff2fce5d7d380a8deb36b5f4851f60a27ed14) in a class without parent, which is, believe it or on, a runtime error:

https://3v4l.org/RVCsd
https://3v4l.org/7E30a

Ocramius:
Nice! Indeed, should be raised as PHP bug

Ocramius:
Argh, just noticed that we'll need a new release of PHPStan before we merge this: https://travis-ci.org/Ocramius/ProxyManager/jobs/203274087#L767-L769

ondrejmirtes:
@Ocramius I don't get it - the message is about PHP-Parser, not PHPStan - if you're using PHP 7.1, you should depend on PHP-Parser 3.x in order for PHPStan to work properly, otherwise you risk parse errors are seen in the build log.  Also - is there a specific reason why you depend on dev version of PHPStan? Because of some bugfix?

I am still undecided how should I approach releasing new versions of PHPStan - I don't want to bother users with BC breaks often so I now prefer to release them in one fell swoop, but it also means that some nice features get to people later than they could.

Ocramius:
Yeah, my bad, need to lock onto newer phpcs. Too bad that php-parser used
unbounded upper constraints 😥.
As for phpstan, I can't merge without a new tag here. Since loads of people
are starting to use it, I'd suggest just releasing a 1.0.0, and simply
being very honest about your intentions for newer versions: major version
bumps aren't to be feared.

src/ProxyManager/Generator/ClassGenerator.php
```diff
@@ -33,7 +33,7 @@ class ClassGenerator extends ZendClassGenerator
     /**
      * {@inheritDoc}
      */
-    public function setExtendedClass($extendedClass) : self
+    public function setExtendedClass($extendedClass) : parent
```
malukenho: Why a `set` method is returning something? seems pretty weird. 

Note: I see the context.

tests/ProxyManagerTest/Factory/RemoteObject/Adapter/BaseAdapterTest.php
```diff
@@ -42,6 +42,7 @@ class BaseAdapterTest extends PHPUnit_Framework_TestCase
      */
     public function testBaseAdapter() : void
     {
+        /* @var $client \Zend\Server\Client|\PHPUnit_Framework_MockObject_MockObject */
         $client = $this
             ->getMockBuilder('Zend\Server\Client')
```
malukenho: `::class`

tests/ProxyManagerTest/Factory/RemoteObject/Adapter/BaseAdapterTest.php
```diff
@@ -76,6 +77,7 @@ public function testBaseAdapter() : void
      */
     public function testBaseAdapterWithServiceMap() : void
     {
+        /* @var $client \Zend\Server\Client|\PHPUnit_Framework_MockObject_MockObject */
         $client = $this
             ->getMockBuilder('Zend\Server\Client')
```
malukenho: same as above

src/ProxyManager/Generator/ClassGenerator.php
Ocramius: That's just the `Zend\Code` API, which is fluent. I can't change the API in a subclass

tests/ProxyManagerTest/Factory/RemoteObject/Adapter/BaseAdapterTest.php
Ocramius: Very good point!

tests/ProxyManagerTest/Factory/RemoteObject/Adapter/BaseAdapterTest.php
Ocramius: 👍

src/ProxyManager/Generator/ClassGenerator.php
malukenho: Yep. I saw the context, just wanted to leave this note 😅

ondrejmirtes:
@Ocramius I just released [PHPStan 0.6.4](https://github.com/phpstan/phpstan/releases/tag/0.6.4) - is it OK for you to use this version in this PR? It contains everything from current master besides BC breaks.

Ocramius:
Unsure if I should just require php-parser in a newer version in dev: we don't really reference it directly, but this silly PHP 7.0 compat is causing me headaches.

Ocramius:
@malukenho if everything looks good, can you please give this a merge?

Ocramius:
@malukenho \o/

PR #375, opened 2017-06-26
Upgrade phpunit dependencies

Ocramius:
This also fixes `master` builds, as phpunit is failing with lowest dependencies due to a subtle bug in how data-providers are reflected


Ocramius:
:ship:

PR #381, opened 2017-07-22
Adding `object` hit based tests and test assets - preparing for PHP 7.2 compatibility

Ocramius:


composer.json
```diff
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "php":                       "^7.1.0",
+        "php":                       "^7.2.0",
```
Majkl578: You should also change Travis to run builds under 7.2 - it's supported as of today.

Ocramius: Oh, nice! I thought `nightly` was all there was to it :)

Majkl578: `nightly` is now 7.3.0-dev. :)

Ocramius:
Aaaand tests pass \o/

Just can't get humbug to run yet, because it needs xdebug :-\

Majkl578:
Also PHPStan fails due to unsupported `object` type - upstream refuses to fix it before 7.2 GA. :( https://github.com/phpstan/phpstan/issues/367

Ocramius:
This requires phpstan `0.9.x` as per https://github.com/phpstan/phpstan/issues/367#issuecomment-329612065

Ocramius:
:ship:

Issue #367, opened 2017-03-14
Property name (suffixes) differ after each generation of a proxy


holtkamp:
**TL;DR;** This is more of a question than an issue: is it really *required* to have a unique suffix of the property names each time they are generated?

Currently each time proxies are generated, the properties are slightly differently named, due to the use of the UniqueIdentifierGenerator (for example for the [$publicProperties](https://github.com/Ocramius/ProxyManager/blob/cce5477857504997baf3168974b8f1283516a686/src/ProxyManager/ProxyGenerator/PropertyGenerator/PublicPropertiesMap.php#L47)). Of course, functionally this is not a problem. But... is it required?

**Rationale**
I know proxies "should" not be committed to a Source Code Management system like GIT / SVN, but PaaS-providers like Heroku 'work' like this (with GIT). Generating the proxies is currently part of our build + deploy process, since a lot (around 900+, 40MB) are involved this is done before pushing the code.

Now on each deploy GIT detects changes in each of the proxies, while actually nothing has changed:

<img width="741" alt="screen shot 2017-03-14 at 10 56 31" src="https://cloud.githubusercontent.com/assets/776405/23896718/a86c449c-08aa-11e7-8e1f-1e63e773de50.png">

Possible options to circumvent this:
- have proxy generation take place after  it has been deployed, when the application starts
  - causes serious delays when starting up the application, significant spikes in response times
- have proxy generation take place as a post-install step in composer
  - at this point the autoload has already dumped
- have proxy generation take place as pre-autoload-dump step in composer
  - viable approach, could consider this
- have proxy generation take place when the code is pushed using [Heroku's 'compile' step](https://devcenter.heroku.com/articles/php-support#custom-compile-step)
  - too vendor specific?
- doubt why the property names change each time...
  - that is why I submitted this issue: why not make the suffix a hash of the classname + propertyName, why **should** it be unique? 
 
A subjective reason to prefer proxy generation before pushing, is the feeling to be 'in-control' about the content being pushed. When a proxy changes, this is clearly visible in the GIT history of Heroku application. Of course, having them change each time, kind of makes this useless.

**In summary**
So I guess my main question is: is it really required to have a unique suffix of the property names each time they are generated, would a hash suffice that only changes if really required?

Ocramius:
> This is more of a question than an issue: is it really required to have a unique suffix of the property names each time they are generated?

[...]

PR #384, opened 2017-07-31
Group generated unique identifiers by name

nicolas-grekas:
This is a micro optim, that may open another one later on:
instead of generating all-time unique property names / methods, this PR allows generating per-prefix unique names. That means that e.g. the "valueHolder" property will have the same name in all generated proxies.

The benefit? Better use of the interned string cache of PHP: instead of filling it with variants of `valueHolder*`, there will be only one of them, the uniquely generated one.

The second benefit to be leveraged later is that this generates exactly same-code for magic/common methods. Which means they could be moved into generated traits, and all proxies could then "use" those traits. This would allow the PHP engine to reuse the same opcode-array for several classes, which means increased performance, by reducing the opcode-array size of the generated proxies (since even with OPcache, opcode-arrays need to be copied from shared to local memory for every request.)

### DIFF
--- a/tests/ProxyManagerTest/ProxyGenerator/PropertyGenerator/AbstractUniquePropertyNameTest.php
+++ b/tests/ProxyManagerTest/ProxyGenerator/PropertyGenerator/AbstractUniquePropertyNameTest.php
@@ -34,7 +34,7 @@ use Zend\Code\Generator\PropertyGenerator;
 abstract class AbstractUniquePropertyNameTest extends PHPUnit_Framework_TestCase
 {
     /**
-     * Verifies that a given property name is unique across two different instantiations of the property
+     * Verifies that a given property name is the same across two different instantiations of the property
      */
     public function testUniqueProperty() : void
     {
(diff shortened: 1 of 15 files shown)


tests/ProxyManagerTest/ProxyGenerator/PropertyGenerator/AbstractUniquePropertyNameTest.php
```diff
@@ -34,15 +34,15 @@
 abstract class AbstractUniquePropertyNameTest extends PHPUnit_Framework_TestCase
 {
     /**
-     * Verifies that a given property name is unique across two different instantiations of the property
+     * Verifies that a given property name is the same across two different instantiations of the property
```
nicolas-grekas: Of course, this invalidates this tests, which now verifies that the name is the same.
Is that an issue?

Ocramius:
@nicolas-grekas the default behavior can be changed without public API changes.
The fact that I use a complex hash + random value is just because developers are clever animals (in a bad way), and they should not tamper with the internals of proxies.
That said, I completely agree that this should be improved. 

@holtkamp reported similar issues at https://github.com/Ocramius/ProxyManager/issues/367 - I think there's a much simpler way to fix all of this.

What we can do, is using `serialize(PackageVersions\Versions::VERSIONS)` as the seed for the suffix hash, then add the name of the property to said hash.

This would remove any runtime randomness, and reduce the complexity of the patch too (we'd just introduce a BC break in the `AbstractUniquePropertyNameTest`, similar to what you just did anyway).

Ocramius:
@nicolas-grekas I'm giving it a shot right now and will open a PR against your branch.

Ocramius:
@nicolas-grekas closing here, moved to #385

PR #385, opened 2017-07-31
#384 #367 deterministic generated property suffixes for generated properties

Ocramius:
Replaces #384
Implements #367 requirements

@nicolas-grekas @holtkamp can you look over this?

Note that it will land only in `2.2.0`, which will require PHP 7.2.

### DIFF
--- a/tests/ProxyManagerTest/ProxyGenerator/PropertyGenerator/AbstractUniquePropertyNameTest.php
+++ b/tests/ProxyManagerTest/ProxyGenerator/PropertyGenerator/AbstractUniquePropertyNameTest.php
@@ -42,7 +42,7 @@ abstract class AbstractUniquePropertyNameTest extends PHPUnit_Framework_TestCase
         $property2 = $this->createProperty();
 
         self::assertSame($property1->getName(), $property1->getName());
-        self::assertEquals($property1->getName(), $property2->getName());
+        self::assertSame($property1->getName(), $property2->getName());
     }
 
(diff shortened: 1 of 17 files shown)


tests/ProxyManagerTest/ProxyGenerator/PropertyGenerator/AbstractUniquePropertyNameTest.php
```diff
@@ -34,15 +34,15 @@
 abstract class AbstractUniquePropertyNameTest extends PHPUnit_Framework_TestCase
 {
     /**
-     * Verifies that a given property name is unique across two different instantiations of the property
+     * Verifies that a given property name is the same across two different instantiations of the property
```
Ocramius: This is obviously a change - possibly to be documented as BC break, even if internal.

nicolas-grekas:
> Note that it will land only in 2.2.0, which will require PHP 7.2.

That's quite unfortunate as it means almost nobody will benefit this in years... Any alternative allowing at least a 7.1?

Ocramius:
@nicolas-grekas if people care about performance as much as you do, they will benefit from it once they upgrade to 7.2 anyway, as that's a much bigger win than this patch anyway.

I'm not merging improvements to past releases.

Ocramius:
Merging as per @nicolas-grekas review.

@holtkamp when you can, check this branch locally to see if it mitigates your issue :-)

PR #386, opened 2017-07-31
Removed licence headers from all single source files

Ocramius:
It is redundant and useless, given the LICENSE file is there anyway.

More space, more I/O, more clutter and more maintenance that isn't really needed.


malukenho:
👍

Ocramius:
:ship:

PR #387, opened 2017-08-23
Correct createProxy PHPDoc

PurpleBooth:
There's a minor incorrect bit of PHPDoc in the NullObjectFactory. The interface states that it only takes an object, however, actually you can pass a string that is a class name and that will work too. This corrects that.


GeeH:
HAPPY BIRTHDAY @Ocramius 

0014e0

Ocramius:
Closing as `invalid`, since this is misuse of the existing API

Ocramius:
J/K, it is OK 👍

Thanks!

PR #389, opened 2017-10-04
Remove header license form files

malukenho:


Ocramius:
@malukenho why `2.1.x`?

PR #392, opened 2017-10-08
PHPStan level 7 analysis

snapshotpl:


src/ProxyManager/Configuration.php
```diff
@@ -44,7 +44,7 @@ class Configuration
     protected $generatorStrategy;
 
     /**
-     * @var callable|null
+     * @var AutoloaderInterface
```
Ocramius: This is `|null`

src/ProxyManager/Configuration.php
```diff
@@ -34,7 +34,7 @@ class Configuration
     protected $proxiesTargetDir;
 
     /**
-     * @var string
+     * @var string|null
```
Ocramius: This is never `null`: see default value

.gitignore
```diff
@@ -8,3 +8,4 @@ clover.xml
 .couscous
 humbuglog.json
 humbuglog.txt
+/nbproject/private/
```
Ocramius: Heh, no, this should be in your global `.gitignore`

Ocramius:
:ship:

PR #393, opened 2017-10-09
remove inheritdoc & body parsing from generated proxies

### LINKED ISSUE #229: Unnecessary doc block reflection
Was doing some profiling of an application that uses this library and noticed that this block of code: https://github.com/Ocramius/ProxyManager/blob/master/src/ProxyManager/Generator/MethodGenerator.php#L67-69 is one of the most expensive in my request cycle.

Upon further inspection, in my particular call stack, I noticed that the parent that calls `MethodGenerator::fromReflection()` is https://github.com/Ocramius/ProxyManager/blob/master/src/ProxyManager/ProxyGenerator/LazyLoadingValueHolder/MethodGenerator/LazyLoadingMethodInterceptor.php#L68 which calls the parent `MethodGenerator::fromReflection()` which does the doc block scanning only for it to be overridden with `{@inheritDoc}` anyway.

My question is, have I missed something or am I right in thinking that in my particular call stack the doc block scanning is pointless? And if so, how can I do away with it so it doesn't impact my performance? Happy to work on a PR to achieve this.

Tobion:
This removes the static `{@inheritDoc}` from the generated proxies and also fixes #229.

This reduces the memory and time overhead. A similar thing has been done in symfony for the generated container. See https://github.com/symfony/symfony/pull/18048 and https://github.com/symfony/symfony/pull/24426

`@inheritdoc` is not useful, especially in proxies that you should not explicitly be aware of anyway.
This way we neither need to generate the phpdoc from the parent definition, nor need to overwrite it with the static inheritdoc afterwards.

### DIFF
--- a/src/ProxyManager/Generator/MethodGenerator.php
+++ b/src/ProxyManager/Generator/MethodGenerator.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace ProxyManager\Generator;
 
+use ReflectionMethod;
 use Zend\Code\Generator\MethodGenerator as ZendMethodGenerator;
+use Zend\Code\Generator\ParameterGenerator;
 use Zend\Code\Reflection\MethodReflection;
 
 /**
@@ -21,10 +23,83 @@ class MethodGenerator extends ZendMethodGenerator
     public static function fromReflection(MethodReflection $reflectionMethod) : self
     {
         /* @var $method self */
-        $method = parent::fromReflection($reflectionMethod);
+        $method = self::fromReflectionWithoutDocBlock($reflectionMethod);
 
         $method->setInterface(false);
 
         return $method;
+    }
+
+    /**
+     * @see \Zend\Code\Generator\MethodGenerator::fromReflection
+     */
+    private static function fromReflectionWithoutDocBlock(MethodReflection $reflectionMethod)
+    {
+        $method         = new static();
+        $declaringClass = $reflectionMethod->getDeclaringClass();
+
+        $method->setSourceContent($reflectionMethod->getContents(false));
+        $method->setSourceDirty(false);
+        $method->setReturnType(self::extractReturnTypeFromMethodReflection($reflectionMethod));
+
+        $method->setFinal($reflectionMethod->isFinal());
+
+        if ($reflectionMethod->isPrivate()) {
+            $method->setVisibility(self::VISIBILITY_PRIVATE);
+        } elseif ($reflectionMethod->isProtected()) {
+            $method->setVisibility(self::VISIBILITY_PROTECTED);
+        } else {
+            $method->setVisibility(self::VISIBILITY_PUBLIC);
+        }
+
+        $method->setInterface($declaringClass->isInterface());
+        $method->setStatic($reflectionMethod->isStatic());
+        $method->setReturnsReference($reflectionMethod->returnsReference());
+        $method->setName($reflectionMethod->getName());
+
+        foreach ($reflectionMethod->getParameters() as $reflectionParameter) {
+            $method->setParameter(ParameterGenerator::fromReflection($reflectionParameter));
+        }
+
+        $method->setBody(static::clearBodyIndention($reflectionMethod->getBody()));
+
+        return $method;
+    }
+
+    /**
+     * @see \Zend\Code\Generator\MethodGenerator::extractReturnTypeFromMethodReflection
+     */
+    private static function extractReturnTypeFromMethodReflection(MethodReflection $methodReflection)
+    {
+        $returnType = method_exists($methodReflection, 'getReturnType')
+            ? $methodReflection->getReturnType()
+            : null;
+
+        if (! $returnType) {
+            return null;
+        }
+
+        if (! method_exists($returnType, 'getName')) {
+            return self::expandLiteralType((string) $returnType, $methodReflection);
+        }
+
+        return ($returnType->allowsNull() ? '?' : '')
+            . self::expandLiteralType($returnType->getName(), $methodReflection);
+    }
+
+    /**
+     * @see \Zend\Code\Generator\MethodGenerator::expandLiteralType
+     */
+    private static function expandLiteralType($literalReturnType, ReflectionMethod $methodReflection)
+    {
+        if ('self' === strtolower($literalReturnType)) {
+            return $methodReflection->getDeclaringClass()->getName();
+        }
+
+        if ('parent' === strtolower($literalReturnType)) {
+            return $methodReflection->getDeclaringClass()->getParentClass()->getName();
+        }
+
+        return $literalReturnType;
     }
 }
--- a/src/ProxyManager/Generator/MethodGenerator.php
+++ b/src/ProxyManager/Generator/MethodGenerator.php
@@ -26,10 +26,7 @@ class MethodGenerator extends ZendMethodGenerator
     {
         $method = new static();
 
-        $method->setSourceContent($reflectionMethod->getContents(false));
-        $method->setSourceDirty(false);
         $method->setReturnType(self::extractReturnTypeFromMethodReflection($reflectionMethod));
-
         $method->setFinal($reflectionMethod->isFinal());
 
         if ($reflectionMethod->isPrivate()) {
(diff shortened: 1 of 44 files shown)


src/ProxyManager/Generator/MethodGenerator.php
```diff
@@ -21,10 +23,85 @@ class MethodGenerator extends ZendMethodGenerator
...
         /* @var $method self */
-        $method = parent::fromReflection($reflectionMethod);
+        $method = self::fromReflectionWithoutDocBlock($reflectionMethod);
 
         $method->setInterface(false);
 
         return $method;
     }
+
+    public static function fromReflectionWithoutDocBlock(MethodReflection $reflectionMethod)
```
Tobion: Those methods are basically copied from the parent with the phpdoc logic removed. The two private methods needed to be copied as well. It's not that nice. Maybe we can ask zend code to add an option for this or provide another extension point.

Ocramius: Isn't it easier to call tge parent method, then set the internal docblock generator?

src/ProxyManager/Generator/MethodGenerator.php
```diff
@@ -21,10 +23,83 @@ class MethodGenerator extends ZendMethodGenerator
...
 
         $method->setInterface(false);
 
         return $method;
     }
+
+    /**
+     * @see \Zend\Code\Generator\MethodGenerator::fromReflection
+     */
+    private static function fromReflectionWithoutDocBlock(MethodReflection $reflectionMethod)
```
Ocramius: Isn't it easier to just introduce a `parent::` call and set the docblock generator in here?

Also, please don't squash - comments were already being hidden because of the squash...

Tobion: I'm not sure what you mean. I could move the whole `fromReflectionWithoutDocBlock` to `fromReflection` directly. But then it's not that abvious what has changed.

Ocramius: @Tobion I'm mainly concerned with extracting logic from `zend-code`. A `removeDocblock()` method on the generator would be best, and it can probably easily be implemented in `zend-code` directly. That's what I'm suggesting, basically: we should use the normal facilities, then remove the docblocks later.

Tobion: If you mean to call the parent and then `$method->setDocBlock(null)`, it would still execute `\Zend\Code\Generator\DocBlockGenerator::fromReflection` for nothing.

Ocramius: >  it would still execute `\Zend\Code\Generator\DocBlockGenerator::fromReflection` for nothing.

Not a problem. This is not performance sensitive code.

Tobion:
I was able to also remove the body parsing which actually needs file IO and tokenizing php code. We can skip that as well since the body of proxies is overwritten anyway.

Tobion:
I did some benchmarks. Generating the proxies in a Symfony project only takes 45% of the time. It saves **300 ms and 360 kB of memory** in a small project with only 6 proxies (12 generation calls because https://github.com/symfony/symfony/blame/3.4/src/Symfony/Bridge/ProxyManager/LazyProxy/PhpDumper/ProxyDumper.php#L126 is executed twice for every definition).

That is quite a saving even in such a small project.

Ocramius:
@Tobion generating proxies is an operation that is supposed to be cached: I'm not interested in improving performance on this side of the logic.

src/ProxyManager/Generator/MethodGenerator.php
```diff
@@ -16,15 +18,74 @@
...
-     * {@inheritDoc}
+     * Similar to fromReflection() but without copying the method body and phpdoc.
+     *
+     * @see \Zend\Code\Generator\MethodGenerator::fromReflection
      */
-    public static function fromReflection(MethodReflection $reflectionMethod) : self
+    public static function fromReflectionWithoutBodyAndDocBlock(MethodReflection $reflectionMethod) : self
     {
-        /* @var $method self */
-        $method = parent::fromReflection($reflectionMethod);
```
Ocramius: As mentioned before, the `parent::fromReflection()` call should stay, while the docblock should be removed later on, once the generator instance is available.

src/ProxyManager/Generator/MethodGenerator.php
```diff
@@ -16,15 +18,74 @@
...
+
+        if (! $returnType) {
+            return null;
+        }
+
+        if (! method_exists($returnType, 'getName')) {
+            return self::expandLiteralType((string) $returnType, $methodReflection);
+        }
+
+        return ($returnType->allowsNull() ? '?' : '')
```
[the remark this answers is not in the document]
Ocramius: All this stuff should just be left to the original lib, and not copied in here.

I already fix bugs in zend-code: don't want to fix them twice.

Ocramius:
@Tobion it is kinda useless to keep working in this direction since all I asked is for the removal of the docblock after the default constructor is removed... :-\

Ocramius:
@Tobion it is kinda useless to keep working in this direction since all I asked for is the removal of the docblock *after* the existing zend-code named constructor is called... :-\

Tobion:
> @Tobion generating proxies is an operation that is supposed to be cached: I'm not interested in improving performance on this side of the logic.

You should consider this not true for development. And that is where OUR time is relevant. Consider that the symfony container is going to be rebuilt more often due to psr4 service discovery, which also needs to track more files for changes. Saving half a second every time, feels worthwhile to me.

I realized `getContent` is also not needed which is also doing IO and tokenization.

The new measurement: Proxy generation only takes 1/4 of the time saving over 400 ms and 420 kB memory for 6 proxies.

Ocramius:
> You should consider this not true for development.

Disable lazy services for development then.

I'm not going to work on improving performance of a pre-cache step, sorry: it is not worth my time, it is not worth your time, and it is generally just a massive pain that leads to shorcuts that are just more wasted maintenance and bugfixing time (look at the code you added - humbug pretty much deals with finding these issues for you).

If you want to improve performance massively, then work on zend-code directly.

I'm also not happy with this performance, but I simply won't go in this direction without a bigger plan of attack.

Even then, I was going to make these steps even slower by moving to an AST-based reflector in the future (slow-down of ~10x).

Ocramius:
That said, the original scope was:

 > remove inheritdoc from generated proxies

If this patch is aiming at that, then all good.

If you are aiming at anything further, please do stop now, because I will not merge it.

src/ProxyManager/Generator/MethodGenerator.php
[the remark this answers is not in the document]
Tobion: Again, if you don't want this code here, we can ask zend-code to add such functionality without parsing body and phpdoc like done here.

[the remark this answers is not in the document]
Ocramius: Yes, that would be precisely the fix to the root cause here. 

`Zend\Code` can move to lazy evaluation of most reflection details, but it needs an overhaul and more work, but that would be the correct endpoint where to work.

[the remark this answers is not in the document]
Ocramius: Probably not with that API. Rather a `::copyMethodSignature()` instead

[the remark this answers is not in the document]
Tobion: The naming makes sense. I'll create a PR.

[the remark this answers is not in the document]
Tobion: https://github.com/zendframework/zend-code/pull/132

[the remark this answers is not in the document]
Ocramius: @Tobion I released zend-code 3.3.0, so this should be un-stuck now :-)

Ocramius:
Merged locally, huge improvement :+1: 

Before:

```
time ./vendor/bin/phpunit
real	0m16.605s
user	0m15.352s
sys	0m1.060s
```

After:
```
time ./vendor/bin/phpunit
real	0m5.045s
user	0m4.584s
sys	0m0.332s
```

Thanks @Tobion, and sorry for the [initial skeptic comments](https://github.com/Ocramius/ProxyManager/pull/393#issuecomment-335573487): patch became much cleaner after https://github.com/zendframework/zend-code/pull/132! \o/

PR #395, opened 2017-11-10
Remove direct reference to value holder property

### LINKED ISSUE #394: Lazy initializers turn proxy value-holder to a reference
A proxy initializer is a closure with this signature:
`function (&$wrappedInstance, \ProxyManager\Proxy\LazyLoadingInterface $proxy [...]);`

This closure is called to initialize a proxy using something like:
`$this->initializer->__invoke($this->valueHolder, $this [...]);`

If for some reason (circular ref, bad initializer, or PHP bug), the `$wrappedInstance` reference is not released, then the `$this->valueHolder` internal property keeps the "is_ref" internal flag "on", meaning eg all clones of the current proxy will share the same `$this->valueHolder` reference.

This is (one of?) the root issue that causes https://github.com/symfony/symfony/issues/23110.

As a work around to confirm the issue, I tried changing the `$instance->initializer = $initializer;` line in the `staticProxyConstructor()` method to that (quick and dirty, suffixes removed):
```php
        $instance->initializer = function ($dummy, ...$args) use ($instance, $initializer) {
            $initializer($value, ...$args);
            $instance->valueHolder = $value;
        };
```

This fixes https://github.com/symfony/symfony/issues/23110.

All versions of the package are affected.

nicolas-grekas:
Fixes #394

### DIFF
--- a/src/ProxyManager/ProxyGenerator/LazyLoadingValueHolder/MethodGenerator/InitializeProxy.php
+++ b/src/ProxyManager/ProxyGenerator/LazyLoadingValueHolder/MethodGenerator/InitializeProxy.php
@@ -39,11 +39,12 @@ class InitializeProxy extends MethodGenerator
         $this->setDocblock('{@inheritDoc}');
 
         $initializer = $initializerProperty->getName();
+        $valueHolder = $valueHolderProperty->getName();
 
         $this->setBody(
-            'return $this->' . $initializer . ' && $this->' . $initializer
-            . '->__invoke($this->' . $valueHolderProperty->getName()
-            . ', $this, \'initializeProxy\', array(), $this->' . $initializer . ');'
+            'return $this->' . $initializer . ' && ($this->' . $initializer
+            . '->__invoke($' . $valueHolder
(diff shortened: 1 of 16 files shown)


nicolas-grekas:
This can be tested by using a closure that forces the reference to be kept.
Eg `$foo->a =& $wrappedValue;` where `$foo` is in the `use` of the initializer.

Ocramius:
Ha, fun fix!

@nicolas-grekas the test still needs to be part of the test suite though. Can you maybe make a `.phpt` for that? See https://github.com/Ocramius/ProxyManager/tree/master/tests/language-feature-scripts

nicolas-grekas:
@Ocramius will do. I submitted this against branch 1.0.x, so that the fix comes to Symfony 2/3 users. Can you confirm this is OK for you?

Ocramius:
@nicolas-grekas `1.0.x` is intentionally no longer maintained unless this is a security issue (not the case): see https://ocramius.github.io/blog/proxy-manager-2-0-0-release/

nicolas-grekas:
So I should resubmit against 2.1, isn't it?

Ocramius:
@nicolas-grekas send it against `master`: I'll backport it for 7.0 and 7.1 versions :+1:

PR #398, opened 2017-11-27
Updated `@var` annotation tags to the standard form

ondrejmirtes:
Something to make upgrading to PHPStan 0.9 easier 😊 One of the BC breaks is that the only `@var` form supported is `@var Type $variableName` and it must be inside a DOC COMMENT token (`/** ... */`).


Ocramius:
Groan, the phpstan dependency needs updating, will likely have to delay until the stable version is out :-P

PR #399, opened 2017-12-04
Use assertDirectoryExists

carusogabriel:

PR #403, opened 2017-12-10
Use Null Coalesce Operator

carusogabriel:


Ocramius:
@carusogabriel was this closed on purpose?

carusogabriel:
@Ocramius Gonna fix those PHPStan errors you mentioned in #399 first, and maybe propose an implementation of a CS. Better than do this changes "just to do", right? :smile:

PR #404, opened 2017-12-11
Fixed #364

### LINKED ISSUE #364: Proxy class can not be written to file, when it was already assembled before
As described here https://github.com/PHP-DI/PHP-DI/issues/459, the configuration to write proxies to a file is not considered when generating the proxy class name (especially the hash of the proxy parameters). 

Not sure it *should* be, but this makes it impossible to write a proxy to a file, when it was already 'loaded' once with a configuration to not write it to a file. 

This typically happens when one wants to generate proxy files in a development environment / as part of a deploy process, where some proxies are already loaded during bootstrapping of the functionality that should actually generate the proxies... (hope that is not too abstract 😄 )

madlexx:
Fixed issue #364. File does not written because autoload executing much earlier than boolean returned.


Ocramius:
Requires a test case

src/ProxyManager/Factory/AbstractBaseFactory.php
```diff
@@ -65,8 +66,27 @@ protected function generateProxy(string $className, array $proxyOptions = []) :
             ->configuration
             ->getClassNameInflector()
             ->getProxyClassName($className, $proxyParameters);
-
-        if (! class_exists($proxyClassName)) {
+        
+        $strategy = $this->configuration->getGeneratorStrategy();                                                  
+                                                                                                                   
+        if ($strategy instanceof FileWriterGeneratorStrategy) {
```
Ocramius: This couples the factory implementation to an implementation of the `GeneratorStrategy` - please don't do this. The only assumption is that you get a generator strategy, but you can't upcast and check it.

madlexx: @Ocramius. Please check, it is updated.

Ocramius:
@madlexx you adapted the broken tests, but didn't introduce a test verifying the issue around #364 :-\

Ocramius:
@madlexx check https://github.com/Ocramius/ProxyManager/tree/master/tests/language-feature-scripts for the absolute simplest way to implement such a test.

You can register an autoloader in a `.phpt` test.

madlexx:
@Ocramius  These tests seem to work.

madlexx:
@Ocramius Please check is this realization is good for you? Thanks.

tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
```diff
@@ -130,4 +133,117 @@ public function testWhenFailingAllTemporaryFilesAreRemoved() : void
...
+                    /* @noinspection UsingInclusionOnceReturnValueInspection */
+                    require_once $tmpFile;
+                }
+
+                return true;
+            }));
+
+        spl_autoload_register(function ($name) use ($fqcn) {
+            if ($name === $fqcn) {
+                throw new \Exception(sprintf("Cannot load class %s", $name));
```
Ocramius: Oh, now I understand what the bug is about. I don't think this needs fixing, tbh - this is a broken autoloader, not a broken ProxyManager detail.

The simplest solution for this is to have the proxy autoloader on top of the autoload stack, but overall the sanest solution is to trash the broken autoloader. Autoloaders should have a `:bool` signature and only crash on fatals...

madlexx: @Ocramius. Thanks for comment. I understand this, but many old projects has "bad" autoloader, and this fix can make them work. In another side if proxies does not generated before  script run we will have error even if the proxy autoloader are registered at top. ZF1 one of "wrong" loader examples. Unfortunately not all projects has composer autoloader or has custom autoloader that followed by best practices.

madlexx: With this patch they will generate proxy if they don't generated for "file write" strategy without the work break.

Ocramius:
Was this supposed to be closed?

PR #405, opened 2017-12-15
Remove non needed requirement

exussum12:
The requirement isn't used and composer install works fine without it


Ocramius:
This dep was needed due to humbug not installing correctly without it. I'll merge this in once I've solved the phpstan issues :)

PR #408, opened 2018-02-02
Add return type `self` to generated method

malukenho:
PS: Zend-Code doesn't support `self` as a return type.

### DIFF
--- a/src/ProxyManager/ProxyGenerator/NullObject/MethodGenerator/StaticProxyConstructor.php
+++ b/src/ProxyManager/ProxyGenerator/NullObject/MethodGenerator/StaticProxyConstructor.php
@@ -35,6 +35,7 @@ class StaticProxyConstructor extends MethodGenerator
             Properties::fromReflectionClass($originalClass)->getPublicProperties()
         );
 
+        $this->setReturnType($originalClass->getName());
         $this->setDocBlock('Constructor for null object initialization');
         $this->setBody(
             'static $reflection;' . "\n\n"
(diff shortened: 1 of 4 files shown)


src/ProxyManager/ProxyGenerator/NullObject/MethodGenerator/StaticProxyConstructor.php
```diff
@@ -35,6 +35,7 @@ function (ReflectionProperty $publicProperty) : string {
             Properties::fromReflectionClass($originalClass)->getPublicProperties()
         );
 
+        $this->setReturnType($originalClass->getName());
```
Ocramius: It should return the proxy class name, not the original class name. Is that feasible, or is that context missing?

malukenho: Well, since the proxy class extends the original class, I think there's no problem with that approach.

Ocramius: That would be inaccurate: can we try using the exact type instead, please?

src/ProxyManager/ProxyGenerator/NullObject/MethodGenerator/StaticProxyConstructor.php
```diff
@@ -35,6 +35,7 @@ function (ReflectionProperty $publicProperty) : string {
             Properties::fromReflectionClass($originalClass)->getPublicProperties()
         );
 
+        $this->setReturnType($this->getName());
```
Ocramius: Ran the tests locally: `$this->getName()` is the method name, not the class name :S

Ocramius:
Manually rebased and applied to `master` as per 06928065654ed6aabdbd5a35cf4c002b74186fa7 (not subsequent commits). Thanks @malukenho!

Ref: f91b540774c4cd56adac4dbd400cbbc01bab6bcc

PR #411, opened 2018-02-10
Deterministic identifiers

lstrojny:
This PR makes identifier generation deterministic. This is one of the puzzle pieces for https://github.com/symfony/symfony/issues/25958.

### DIFF
--- a/src/ProxyManager/Generator/Util/UniqueIdentifierGenerator.php
+++ b/src/ProxyManager/Generator/Util/UniqueIdentifierGenerator.php
@@ -41,15 +41,8 @@ abstract class UniqueIdentifierGenerator
      */
     public static function getIdentifier(string $name) : string
     {
-        return str_replace(
-            '.',
-            '',
-            uniqid(
-                preg_match(static::VALID_IDENTIFIER_FORMAT, $name)
-                ? $name
-                : static::DEFAULT_IDENTIFIER,
-                true
-            )
-        );
+        $identifier = preg_match(static::VALID_IDENTIFIER_FORMAT, $name) ? $name : static::DEFAULT_IDENTIFIER;
+
+        return $identifier . substr(hash('sha256', $identifier), 0, 14);
     }
 }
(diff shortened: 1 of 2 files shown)


Ocramius:
Hmm, I thought I already did this in the past...

I'm fine with using deterministic values, but this approach isn't deterministic either. Could you maybe use a short md5 (4 chars or so) from the md5 of the given name instead?

Ocramius:
I'd also need to check if that still works, as unique names are needed within proxies....

A test about idempotence of this API is also needed

lstrojny:
Right now `ProxyManagerTest\Generator\Util\UniqueIdentifierGeneratorTest::testGeneratesUniqueIdentifiers()` looks like idempotence is specifically not the plan. I've submitted a changed version anyway. Let me know what you think.

src/ProxyManager/Generator/Util/UniqueIdentifierGenerator.php
```diff
@@ -41,15 +41,8 @@
...
-            uniqid(
-                preg_match(static::VALID_IDENTIFIER_FORMAT, $name)
-                ? $name
-                : static::DEFAULT_IDENTIFIER,
-                true
-            )
-        );
+        $identifier = preg_match(static::VALID_IDENTIFIER_FORMAT, $name) ? $name : static::DEFAULT_IDENTIFIER;
+
+        return $identifier . substr(hash('sha256', $identifier), 0, 8);
```
Ocramius: Note that no security constraints apply here, so I don't think we need this, still good improvement 👍

lstrojny: I know but using MD5 anywhere will just set a bad precedent. It’s easier to explain "never use MD5 for anything" 😄

lstrojny:
Rebased against 2.1, made the identifiers slightly longer (as the test asks for).

Tobion:
This has already been fixed in master in #385 but 2.2 release requires php 7.2. This is probably why @lstrojny does not use it yet in his tests.

Ocramius:
Ah, I see... Yeah, that's not going to go to `2.x`...

lstrojny:
@Ocramius hah, good to know. Any chance to get this backported to a PHP 7.1 compatible version?

Ocramius:
Nope, no chance - old versions are only for quite critical bugfixes, sorry.

stof:
then, shouldn't this PR be closed then ?

Ocramius:
I didn't yet check if `3.x` matches

Ocramius:
Closing as duplicate of https://github.com/Ocramius/ProxyManager/pull/385

PR #412, opened 2018-03-19
FileWriterGenerator temp name fix

mainegreen:
tempname will not work on location that includes proxy php class file name, as passed  in location. This is generated in generate() as from the $this->fileLocator->getProxyFileName($className). In reality we need the dirname base path of the desired location to pass into tempname.


Ocramius:
@mainegreen this requires a test case demonstrating how the current approach is incorrect/fails, and that the patch fixes that.

fmasa:
I just noticed this PR :man_facepalming: I submitted PR with both fix and test change demonstrating the issue in https://github.com/Ocramius/ProxyManager/pull/438

PR #413, opened 2018-05-08
Document writing generating proxies to files

2Dsharp:
Added documentation about generating the proxies and storing them as files before trying to autoload them from the cache directory.

### DIFF
--- a/docs/tuning-for-production.md
+++ b/docs/tuning-for-production.md
@@ -6,35 +6,35 @@ title: Tuning the ProxyManager for production
 
 By default, all proxy factories generate the required proxy classes at runtime.
 
-To generate proxies and store them as files, you need to use the `FileWriterGeneratorStrategy` by configuring ProxyManager.
-
-```php
-    
-$config = new \ProxyManager\Configuration();
-
-$fileLocator = new \ProxyManager\FileLocator\FileLocator(__DIR__.'/my/generated/classes/cache/dir');
-$config->setGeneratorStrategy(new \ProxyManager\GeneratorStrategy\FileWriterGeneratorStrategy($fileLocator));
-
-// pass the configuration to proxymanager factory
-$factory = new ProxyManager\Factory\LazyLoadingValueHolderFactory($config);
-```
-The files generated in the directory will be needed to autoload them.
-
 Proxy generation causes I/O operations and uses a lot of reflection, so be sure to have
 generated all of your proxies **before deploying your code on a live system**, or you
 may experience poor performance.
+
+To generate proxies and store them as files, you need to use the `FileWriterGeneratorStrategy` 
+by configuring ProxyManager.
+The files generated in the directory will be needed to autoload them.
 
 You can configure ProxyManager so that it will try autoloading the proxies first.
 Generating them "bulk" is not yet implemented:
 
 ```php
+    
 $config = new \ProxyManager\Configuration();
+
+// generate the proxies and store them as files
+$fileLocator = new \ProxyManager\FileLocator\FileLocator(__DIR__.'/my/generated/classes/cache/dir');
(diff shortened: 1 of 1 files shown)


docs/tuning-for-production.md
```diff
@@ -6,6 +6,20 @@ title: Tuning the ProxyManager for production
 
 By default, all proxy factories generate the required proxy classes at runtime.
 
+To generate proxies and store them as files, you need to use `FileWriterGeneratorStrategy` by passing it to the configuration before setting up other factories:
```
Ocramius: /to use/to use the

Ocramius: `the configuration` is unclear. Maybe just saying `configuring ProxyManager`?

docs/tuning-for-production.md
```diff
@@ -6,6 +6,20 @@ title: Tuning the ProxyManager for production
...
+    
+$config = new \ProxyManager\Configuration();
+
+$fileLocator = new \ProxyManager\FileLocator\FileLocator(__DIR__.'/my/generated/classes/cache/dir');
+$config->setGeneratorStrategy(new \ProxyManager\GeneratorStrategy\FileWriterGeneratorStrategy($fileLocator));
+
+// pass the configuration to proxymanager factory
+$factory = new ProxyManager\Factory\LazyLoadingValueHolderFactory($config);
+```
+This will write the generated proxies in a directory which can be used directly instead of generating them again and again.
```
Ocramius: > in a directory which can be used directly instead of generating them again and again.

The `FileWriterGeneratorStrategy` writes files regardless. What we are achieving here is that the files are being written at all, which is needed for us to make them part of our deployable. Can you try adjusting the wording according to that?

docs/tuning-for-production.md
```diff
@@ -6,6 +6,20 @@ title: Tuning the ProxyManager for production
...
+
+```php
+    
+$config = new \ProxyManager\Configuration();
+
+$fileLocator = new \ProxyManager\FileLocator\FileLocator(__DIR__.'/my/generated/classes/cache/dir');
+$config->setGeneratorStrategy(new \ProxyManager\GeneratorStrategy\FileWriterGeneratorStrategy($fileLocator));
+
+// pass the configuration to proxymanager factory
+$factory = new ProxyManager\Factory\LazyLoadingValueHolderFactory($config);
```
Ocramius: We may need a full example with this combined with the autoloader.

docs/tuning-for-production.md
2Dsharp: Not sure what your second comment means. Could you clarify?

Ocramius: `the configuration` refers to something specific, yet you don't name it here. I suggest removing `by passing it to the configuration before setting up other factories:` and having `by configuring ProxyManager:`

docs/tuning-for-production.md
2Dsharp: Is `$config->setProxiesTargetDir()` absolutely necessary for generating the proxies? My implementation generates and uses the proxies from Auryn, so it needs that line otherwise it will throw a `ReflectionException`

2Dsharp: The two examples can be put together as one. Is that an option?

Ocramius: @2Dsharp yes, making a single example is a good idea.

As for `$config->setProxiesTargetDir()`, it is not strictly needed, but you cannot rely on `ProxyManager\Configuration` to build correct instances without it...

For example, the `FileLocator` is already correctly configured if `setProxiesTargetDir()` was used.

2Dsharp: I am merging the two examples into one.

PR #416, opened 2018-05-12
Tidied up the docs and simply removed weasel words

GeeH:


Majkl578:
I don't understand the changes to line breaks.

GeeH:
For consistency I've made the linebreaks happen at 120 chars which I've done across Marco's other projects in the past.

Majkl578:
Tha doesn't make sense. It makes diffs in GitHub look weird (see screenshot above). I recently decided to not wrap full sentectes at all and leave that to the GH diff or IDE, it's the most convenient and natural way for text documents.

Majkl578:
Btw such line breaks make changes to the document much harder to review, often requiring changing whole paragraph.

GeeH:
I don't agree with you. @Ocramius  asked for this PR and in my opinion, consistent line breaks are an important part of markdown formatting. In the same way that you wouldn't let your IDE add soft line breaks in your code, you shouldn't do it in your markdown either. It's unfortunate that the review is more difficult because I've changed the line breaks, but that's part of the PR so it will need to be reviewed as is.

Majkl578:
> consistent line breaks are an important part of markdown formatting

I disagree.

> In the same way that you wouldn't let your IDE add soft line breaks in your code

I disagree with this statement as well. PHP code has no hard limit (PSR-2 even forbids having hard limit).

GeeH:
I still disagree. If you want to reformat before merging feel free.

Ocramius:
I'm personally for wrapping at around 80 chars, as 90%+ of the time when I read documentation, I read it from the sources with a no-wrap editor anyway, and with markdown rendering off (because markdown loves to skip details).

Therefore keeping the wrapping here: it is an improvement from my PoV.

docs/credits.md
```diff
@@ -24,3 +25,7 @@ The idea was originated by a [talk about Proxies in PHP OOP](http://marco-pivett
 - [Aleksey Khudyakov](https://github.com/Xerkus)
 - [Alexander](https://github.com/asm89)
 - [Raul Fraile](https://github.com/raulfraile)
+
+### Documentation
+
+[Gary Hockin](https://twitter.com/GeeH) is amazing and spent literally 20 minutes tidying up the documentation.
```
Ocramius: :+1:

docs/credits.md
```diff
@@ -4,7 +4,8 @@ title: Credits
 
 # Credits
 
-The idea was originated by a [talk about Proxies in PHP OOP](http://marco-pivetta.com/proxy-pattern-in-php/) that I gave at the [@phpugffm](https://twitter.com/phpugffm) in January 2013.
+The idea was originated by a [talk about Proxies in PHP OOP](http://marco-pivetta.com/proxy-pattern-in-php/) that I gave at
```
Ocramius: Dang, this is now offline...

This is at https://ocramius.github.io/presentations/proxy-pattern-in-php/ now

Ocramius:
Thanks @GeeH :ship:

PR #419, opened 2018-06-25
Fix proxying an interface and using a dynamic property on the proxy

nicolas-grekas:
Submitting on branch 2.1 because that's the lowest one compatible with PHP 7.1, thus Symfony 4.

When an interface is proxied, should the dynamic property be forwarded to the wrapped value? ~I would prefer not actually, so it'd stay only on the proxy.~

### DIFF
--- a/src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
+++ b/src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
@@ -49,6 +49,7 @@ class PublicScopeSimulator
      *                                              to read the property. `$this` if none provided
      * @param string|null       $returnPropertyName name of the property to which we want to assign the result of
      *                                              the operation. Return directly if none provided
+     * @param string|null       $interfaceName      name of the proxified interface if any
      *
      * @return string
      *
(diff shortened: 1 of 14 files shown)


nicolas-grekas:
Now updated with a fix. Property is forwarded on the value-holder, should play better with interfaces that have semantics related to dynamic props (if that makes sense.)

tests/language-feature-scripts/lazy-loading-value-holder-interface-proxy.phpt
```diff
@@ -0,0 +1,36 @@
...
+
+$proxy = $factory
+    ->createProxy(MyInterface::class, function (& $wrapped, $proxy, $method, array $parameters, & $initializer) {
+        $initializer = null;
+        $wrapped     = new MyClass();
+    });
+
+$proxy->do();
+$proxy->someDynamicProperty = ' World';
+echo @$proxy->someDynamicProperty;
```
Ocramius: Could you remove the suppression and assert that there is a notice here? You can use regular expressions when using EXPECTF if I remember correctly.

src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
```diff
@@ -96,23 +104,29 @@ public static function getPublicAccessSimulationCode(
...
             return '';
         }
 
-        return '    $backtrace = debug_backtrace(false);' . "\n"
+        $code = '    $backtrace = debug_backtrace(false, 1);' . "\n"
             . '    trigger_error(' . "\n"
             . '        sprintf(' . "\n"
             . '            \'Undefined property: %s::$%s in %s on line %s\',' . "\n"
-            . '            get_parent_class($this),' . "\n"
+            . '            ' . (null !== $interfaceName ? var_export($interfaceName, true) : 'get_parent_class($this)') . ',' . "\n"
```
Ocramius: Would it make sense to pass in the parent symbol name, instead of relying on `get_parent_class()` at runtime at all?

PR #421, opened 2018-08-17
FileWriterGeneratorStrategy: Set file permissions on the generated proxy

hultberg:
The exact call for chmod is inspired by twig: https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Cache/Filesystem.php#L64

### DIFF
--- a/tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
+++ b/tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
@@ -45,6 +45,13 @@ class FileWriterGeneratorStrategyTest extends TestCase
         self::assertGreaterThan(0, strpos($body, $className));
         self::assertFalse(class_exists($fqcn, false));
         self::assertFileExists($tmpFile);
+        self::assertFileIsReadable($tmpFile);
+
+        // a user not on php.net recommended calling this as we have just called chmod on a file.
+        clearstatcache();
+
+        $perm = 0666 & ~umask();
+        self::assertTrue($perm === (fileperms($tmpFile) & 0644), 'File permission was not correct: ' . decoct($perm));
 
         /* @noinspection PhpIncludeInspection */
         require $tmpFile;
--- a/tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
+++ b/tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
@@ -47,7 +47,7 @@ class FileWriterGeneratorStrategyTest extends TestCase
         self::assertFileExists($tmpFile);
         self::assertFileIsReadable($tmpFile);
 
-        // a user not on php.net recommended calling this as we have just called chmod on a file.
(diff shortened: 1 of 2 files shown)


Ocramius:
Requires test case additions. Also, the `chmod` should not be suppressed.

Note that the exact permissions must be verified, or else mutation tests will fail.

hultberg:
Also, is it possible to have a new release of 2.1 series with this fix?

Ocramius:
Yeah, I need to migrate the tests to use infection/infection - didn't yet
find the time to do it

hultberg:
Opened #422 for 2.1.x

Added a simple test for the file permission.

Ocramius:
Unless this is a security fix, this should only be in the latest release

tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
```diff
@@ -45,6 +45,13 @@ public function testGenerate() : void
         self::assertGreaterThan(0, strpos($body, $className));
         self::assertFalse(class_exists($fqcn, false));
         self::assertFileExists($tmpFile);
+        self::assertFileIsReadable($tmpFile);
+
+        // a user not on php.net recommended calling this as we have just called chmod on a file.
+        clearstatcache();
+
+        $perm = 0666 & ~umask();
+        self::assertTrue($perm === (fileperms($tmpFile) & 0644), 'File permission was not correct: ' . decoct($perm));
```
Ocramius: `assertSame` here please - also some comments on the `$perm` needed. Need to also make sure this test is cross-platform compliant: what happens on windoze for example?

hultberg: @Ocramius Improved

hultberg:
A friendly bump. :smiley:

Ocramius:
Will try to work on this library over the weekend, while at https://www.froscon.de/

Ocramius:
After checking this, I will only move it into `master` (`3.x`):

 1. the `FileWriterGeneratorStrategy` is to be configured manually by consumers, and should be used in with paths only accessible to the running application (generally the case in VMs/containers, obviously less so for shared hosting environments)
 2. this change is a potential BC break (will mark it as such)
 3. counts as a new feature

Ocramius:
Manually merged to `master` after a rebase, thanks!

Ocramius:
This test fails locally on `master` - will need to revert it.

Ocramius:
Hmm, no, only fails in the context of the IDE, probably because the IDE plays around with `umask()`:

```
File permission was not correct: 664
Failed asserting that 420 is identical to 436.
```

@hultberg do you also develop on PHPStorm? Able to check this?

Ocramius:
Yep, this assertion is incorrect, as it assumes a `umask()` set to `022`:

```php
        self::assertSame($perm, fileperms($tmpFile) & 0644, 'File permission was not correct: ' . decoct($perm));
```

Reverting the entire patch and re-opening here.

hultberg:
@Ocramius Sorry for the late response. I did in fact use phpstorm and ran the tests via my terminal and not phpstorm itself. Glad to see you resolved the issues. :smile:

PR #422, opened 2018-08-17
[2.1.x] FileWriterGeneratorStrategy: Set file permissions when writing proxy file

hultberg:
Ported for 2.1.x


Ocramius:
I'll do a backport after the original patch is done

hultberg:
@Ocramius Alright, I'll close this one then. Thanks

Ocramius:
After careful consideration, this is not going to be backported to `2.1.x`, as the `FileWriterGeneratorStrategy` is to be manually configured to write to a location that is under the developer's (only) control.

PR #423, opened 2018-08-25
Typed properties (PHP 7.4, hopefully) example support

Ocramius:


Ocramius:
Ref: https://github.com/php/php-src/pull/3313

src/ProxyManager/ProxyGenerator/Util/Properties.php
```diff
@@ -64,6 +64,27 @@ public function filter(array $excludedProperties) : self
...
+            }
+
+            /** @var $type \ReflectionType */
+            $type = $property->getType();
+
+            if ($type->allowsNull()) {
+                return true;
+            }
+
+            return isset($property->getDeclaringClass()->getDefaultProperties()[$property->getName()]);
```
malukenho:
```php
return $property->getType()->allowsNull()
    || isset($property->getDeclaringClass()->getDefaultProperties()[$property->getName()]);
```

composer.json
```diff
@@ -29,11 +29,7 @@
         "squizlabs/php_codesniffer":    "^2.9.1",
         "couscous/couscous":            "^1.6.1",
         "phpbench/phpbench":            "^0.12.2",
-        "phpstan/phpstan":              "dev-master#856eb10a81c1d27c701a83f167dc870fd8f4236a as 0.9.999",
-        "phpstan/phpstan-phpunit":      "dev-master#5629c0a1f4a9c417cb1077cf6693ad9753895761",
-        "nikic/php-parser":             "^3.1.1",
-        "humbug/humbug":                "1.0.0-RC.0@RC",
-        "padraic/phpunit-accelerator":  "dev-master@DEV"
```
malukenho: Shouldn't it be reverted?

tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
```diff
@@ -4,8 +4,8 @@
 
 namespace ProxyManagerTest\Functional;
 
-use PHPUnit_Framework_MockObject_MockObject as Mock;
 use PHPUnit\Framework\TestCase;
+use PHPUnit_Framework_MockObject_MockObject as Mock;
```
malukenho: As far as I know, we can use the namespaced version here.

tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
```diff
@@ -664,6 +665,34 @@ function ($proxy, $method, $params, & $initializer, array $properties) {
         }
     }
 
+    public function testByRefInitializationOfTypedProperties() : void
+    {
+        $proxyName = $this->generateProxy(ClassWithMixedTypedProperties::class);
+        /* @var $proxy ClassWithMixedTypedProperties */
```
malukenho: use `/**` instead of `/*`

tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
```diff
@@ -746,6 +775,7 @@ private function createInitializer(string $className, $realInstance, Mock $initi
                     self::isInstanceOf(GhostObjectInterface::class),
                     self::isInstanceOf($className)
                 ));
+            ;
```
malukenho: Is it a mistake?

composer.json
Ocramius: I fixed these in `master`: a rebase will remove them

src/ProxyManager/ProxyGenerator/Util/Properties.php
Ocramius: If there's no type, that will crash

Ocramius: Isn't that the same as the current code? Note that `$type` needs to be checked for nullability

tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
Ocramius: 👍

tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
Ocramius: Yes, automated CS added this, I think

src/ProxyManager/ProxyGenerator/Util/Properties.php
malukenho: We could do the unmaintainable version:

```php
return false === $property->hasType()
    || $property->getType()->allowsNull()
    || isset($property->getDeclaringClass()->getDefaultProperties()[$property->getName()]);
```

Ocramius: That wouldn't pass static analysis :-)

tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
```diff
@@ -1353,4 +1387,14 @@ function (VoidCounter $proxy, $method, $params, & $initializer, array $props) us
 
         self::assertSame($initialCounter + $increment, $proxy->counter);
     }
+
+    private static function isPropertyInitialized(object $object, ReflectionProperty $property) : bool
+    {
+        return array_key_exists(
```
nikic: You should be able to use `$property->isInitialized($object)` here.

Ocramius: Does that also consider the `unset()` scenario?

nikic: Yes.

nikic: Hm actually, is this one called on an object with lazy initialization __get? Because I think isInitialized() is going to invoke that.

tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
Ocramius: Checked: will require a second patch targeting PHPUnit upgrade

composer.json
Ocramius: Rebased

src/ProxyManager/ProxyGenerator/Util/Properties.php
Ocramius: This has been tested and changed, since I had to spawn https://bugs.php.net/bug.php?id=77673

tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
Ocramius: Change applied

Ocramius:
Since this is :green_heart: (except for infection, but that's not important for now), :ship: before I get more drinks in.

tests/ProxyManagerTest/Autoloader/AutoloaderTest.php
```diff
@@ -5,6 +5,7 @@
 namespace ProxyManagerTest\Autoloader;
 
 use PHPUnit\Framework\TestCase;
+use PHPUnit_Framework_MockObject_MockObject;
```
sebastianbergmann: PHPUnit 8 removed the BC fallback for `PHPUnit_Framework_MockObject_MockObject` and now only `PHPUnit\Framework\MockObject\MockObject` works (which has been around for a while).

Ocramius: Yes, this patch is old, and repository ia still locked onto phpunit 7, since I rely on reflection-based assertions

sebastianbergmann: Sorry, for not making this clear: PHPUnit already has `PHPUnit\Framework\MockObject\MockObject` and only provides `PHPUnit_Framework_MockObject_MockObject` for BC.

Ocramius: Ah, yup, #441 is on the way

PR #427, opened 2018-09-03
Crashing test assets for magic methods with types/void

Majkl578:
Right now just failing (crashing) test assets.

Related to #350.


Ocramius:
Thanks!

Ocramius:
I wrote a possible (example of a) fix t #457 - not sure if worth pursuing for anything other than `__set()` though

Issue #431, opened 2018-09-22
Fix current escaped mutants


Ocramius:
See https://github.com/infection/infection/issues/486#issuecomment-423764688

PR #433, opened 2018-09-27
Fix passing of variadic arguments to RPC adapters in remote object proxies

Ocramius:


Ocramius:
Build failure is related to https://github.com/Ocramius/ProxyManager/issues/431, not critical for now.

PR #434, opened 2018-09-27
Fix passing of variadic arguments to RPC adapters in remote object proxies (Backport of #443)

Ocramius:
Fixes #443 in 2.2.x


Ocramius:
Will have to ship this as red - build failures are related to build dependencies that aren't there

PR #438, opened 2018-11-18
Fixed location of FileWriterGeneratorStrategy temp file

fmasa:
Hi,
I encountered problem with this (awesome) package on project with configured `open_basedir`.

There seems to be bug in `FileWriterGeneratorStrategy` - file path is passed to `tempnam` instead of directory.
https://github.com/Ocramius/ProxyManager/blob/8d0fa8e3e11cdadaaf4c41732366b5ed48bb0aed/src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php#L77

This results in directory always falling back to `/tmp`.

There is test which should test this, but it's false positive (exception is thrown because new file cannot be created in target directory, temp file is created in /tmp)
https://github.com/Ocramius/ProxyManager/blob/8d0fa8e3e11cdadaaf4c41732366b5ed48bb0aed/tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php#L73

This PR provides fix for the issue.


src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
```diff
@@ -74,9 +75,12 @@ public function generate(ClassGenerator $classGenerator) : string
      */
     private function writeFile(string $source, string $location) : void
     {
-        $tmpFileName = tempnam($location, 'temporaryProxyManagerFile');
+        $directory   = dirname($location);
```
malukenho: In this case, we need to check if the `$location` is a directory first, otherwise, we will move up one directory.

fmasa: `$location` cannot be directory because it's value returned from `FileLocatorInterface::locate()`, which is supposed to return "full path for a given file name".

fmasa: See https://github.com/Ocramius/ProxyManager/pull/438/files/67ddd7b00c70f40cb8469a6bf4e474da44189e95#diff-f7ba53e97e34edc37d8bc3477bea48f2R61

fmasa:
@malukenho I fixed coding standard error that I introduced. But build still fails for unrelated stuff (PHPstan check of different file and mutation testing all over the place). Is there anything else that needs to be done?

src/ProxyManager/Exception/FileNotWritableException.php
```diff
@@ -22,4 +22,13 @@ public static function fromInvalidMoveOperation(string $fromPath, string $toPath
             $toPath
         ));
     }
+
+    public static function fromNotWritableDirectory(string $directory) : self
```
Ocramius: While annoying, please write one test that verifies this method (yes, I also do test these)

tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
```diff
@@ -27,17 +30,32 @@
...
+
+    protected function setUp() : void
+    {
+        $this->tempDir = sys_get_temp_dir() . '/' . self::class;
+
+        if (! is_dir($this->tempDir)) {
+            mkdir($this->tempDir);
+        }
+
+        ini_set('open_basedir', __DIR__ . '/../../..' . PATH_SEPARATOR . $this->tempDir);
```
Ocramius: Is an extension in `require-dev` needed here?

src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
```diff
@@ -74,9 +73,12 @@ public function generate(ClassGenerator $classGenerator) : string
...
+        $tmpFileName = tempnam($directory, 'temporaryProxyManagerFile');
 
-        assert(is_string($tmpFileName));
+        if ($tmpFileName === false) {
+            throw FileNotWritableException::fromNotWritableDirectory($directory);
```
Ocramius: Is there a way to hit this code location?

tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
```diff
@@ -27,17 +30,32 @@
...
  * Note: this test generates temporary files that are not deleted
  */
 class FileWriterGeneratorStrategyTest extends TestCase
 {
+    /** @var string */
+    private $tempDir;
+
+    protected function setUp() : void
+    {
+        $this->tempDir = sys_get_temp_dir() . '/' . self::class;
```
Ocramius: Consider using `uniqid()` or similar, instead of using the class name (which, by the way, includes backslashes: won't work on windows)

tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
Ocramius: What is thia for, precisely?

fmasa: > Is an extension in require-dev needed here?

`open_basedir` is part of core.

> What is thia for, precisely?

This is for testing this scenario. `open_basedir` now restricts access to 2 paths:
- repository root
- specific directory in system temp

> If the directory does not exist or is not writable, **tempnam()** may generate a file in the system's temporary directory, and return the full path to that file, including its name. 

As briefly mentioned in [documentation of tempnam](http://php.net/manual/en/function.tempnam.php), when passed directory is not writable, this function attempts to use system temp. When system temp directory is not writable, file is not created and `false` is returned.

There is alternative approach using `putenv('TMPDIR=/not-writable-location')` as `TMPDIR` env variable can override systemp temp directory for current script. No matter which approach will be used, this line should be probably moved to `testGenerateWillFailIfTmpFileCannotBeWrittenToDisk()` to make it more obivous.

src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
fmasa: Yes, see https://github.com/Ocramius/ProxyManager/pull/438#discussion_r241368819

tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
fmasa: Right, used `uniqid()` instead.

src/ProxyManager/Exception/FileNotWritableException.php
fmasa: :heavy_check_mark: Tested in https://github.com/Ocramius/ProxyManager/pull/438/commits/2414974aad1df2391e629cf1628ce38bbd6d62bd

fmasa:
@Ocramius Is there anything else to be done?

Ocramius:
Note: release may take a while more, as I first need to bring `master` into a decent state again :-\

PR #444, opened 2019-02-27
Run `infection/infection` with `phpdbg`

Ocramius:


Ocramius:
only *one* green though - fixing the other now :-)

PR #446, opened 2019-02-27
Exclude `zendframework/zend-stdlib < 3.2.1` (not compatible with PHP 7.3)

Ocramius:


Ocramius:
No luck yet (something still broken in @phpstan with `--prefer-lowest`) but shipping as-is, since it still got us past the unit tests

PR #448, opened 2019-02-27
Forcing latest `nikic/php-parser` releases for dev dependencies

Ocramius:
Without this dependency, static analysis fails due to PHP 7.2+ incompatibilities


Ocramius:
Green build! :ship:

PR #465, opened 2019-05-23
Introducing initial tentative stub for #463: psalm-specific type declarations

### LINKED ISSUE #463: Provide `@psalm-param class-string<T>` documentation for proxy factories and anything type-hinted `object` and `: object`
See https://psalm.dev/docs/templated_annotations/#param-class-stringlttgt

Ocramius:
Fixes #463

This is an incomplete patch that just serves to display that types can be
improved massively in here, making even this magic contraption of a library
usable downstream with some decent type constraints.

TODOs:

 - [x] need to implement CI steps to ensure type inference works as expected
 - [x] full type coverage in ProxyManager internals (note: no tests yet - those require more work)
 - [x] pin to stable `vimeo/psalm` (`^3.4.0`)
 - [x] split type check test file into multiple type-checked examples
 - [x] raise psalm reporting levels
 - [x] add utilities for PHPStorm type inference

Blocked by:

 - [x] https://github.com/vimeo/psalm/issues/1673
 - [x] https://github.com/vimeo/psalm/issues/1674
 - [x] https://github.com/vimeo/psalm/issues/1675
 - [x] https://github.com/vimeo/psalm/issues/1677
 - [x] https://github.com/vimeo/psalm/issues/1679
 - [x] https://github.com/vimeo/psalm/issues/1680
 - [x] https://github.com/vimeo/psalm/issues/1681
 - [x] https://github.com/vimeo/psalm/issues/1682
 - [x] https://github.com/vimeo/psalm/issues/1683
 - [x] https://github.com/vimeo/psalm/issues/1722
 - [x] https://github.com/vimeo/psalm/pull/1726


composer.json
```diff
@@ -38,7 +38,8 @@
         "phpstan/phpstan":              "^0.11.4",
         "phpstan/phpstan-phpunit":      "^0.11",
         "infection/infection":          "^0.12.2",
-        "symfony/console":              "^4.2.4"
+        "symfony/console":              "^4.2.4",
+        "vimeo/psalm": "^3.3"
```
Ocramius: Note: I just committed whatever I had stashed locally, so I'd just need a rough high-level review of this feature.

type-example.php
```diff
@@ -0,0 +1,84 @@
...
+        $initializer   = null; // disable initialization
+        $wrappedObject = new MyProxiedClass();
+
+        return true;
+    })
+    ->sayHello();
+
+echo (new NullObjectFactory())
+    ->createProxy(MyProxiedClass::class)
+    ->sayHello();
```
Ocramius: This type check seems to fail: `NullObjectFactory#createProxy()` may be annotated incorrectly.

type-example.php
```diff
@@ -0,0 +1,84 @@
...
+        $initializer = function (
+            GhostObjectInterface $proxy,
+            string $method,
+            array $parameters,
+            & $initializer,
+            array $properties
+        ) {
+            $initializer = null; // disable initialization
+        })
+    ->sayHello();
```
Ocramius: This type declaration seems to fail type checks:

```
ERROR: UndefinedInterfaceMethod - type-example.php:49:7 - Method ProxyManager\Proxy\GhostObjectInterface::sayhello does not exist
    ->sayHello();
```

@muglug I'm sorry to chase you across repositories, but do you maybe know whether this is a docblock parser bug? This is the docblock:

https://github.com/Ocramius/ProxyManager/blob/300973221ee90f0f126e5f3b2753410c7c5c88d4/src/ProxyManager/Factory/LazyLoadingGhostFactory.php#L38-L88

muglug: that docblock should be
```php
/**
 * @psalm-template RealObjectType 
 * 
 * @psalm-param class-string<RealObjectType> $className 
 * @psalm-param array{skippedProperties?: array<int, string>} $proxyOptions
 * 
 * @psalm-return RealObjectType&GhostObjectInterface 
 */ 
```

Ocramius: Daaang, so it is just that `?`!

muglug: And missing `$proxyOptions`

stof:
a PR fixing itself ? That's inception.

Ocramius:
> a PR fixing itself ?

What do you mean by that? I didn't catch the joke XD

src/ProxyManager/Inflector/ClassNameInflector.php
```diff
@@ -37,14 +38,14 @@ public function getUserClassName(string $className) : string
         $className = ltrim($className, '\\');
         $position  = strrpos($className, $this->proxyMarker);
 
-        if ($position === false) {
+        if (! is_int($position)) {
```
stof: is this change actually needed by psalm ? It should be able to detect that the other case has an int when comparing to false, as the signature is `int|false`.

stof:
The description contains this:

> Fixes #465

This reference is the PR itself.

Ocramius:
Ha, fixed, thanks!

src/ProxyManager/Inflector/ClassNameInflector.php
Ocramius: The inference for now is `int|bool`, not `int|false`. I'm fine with adapting the code for these scenarios for now, as upstream work for PHPStorm stubs is ongoing.

stof: oh, so PHPStan has the `int|false` type for `strrpos`, but psalm does not have it yet ?

muglug: Psalm definitely has it - @Ocramius is maybe talking about PHPStorm?

Ocramius: Well, this has been inferred as `int|bool` here :|

muglug: Maybe something else is happening? https://psalm.dev/r/d029859bdf works as expected

muglug: Or possibly this was an  attempt to fix the possibly-false operandi issue (that you fixed with an explicit `(int)` cast below

Ocramius:
```
Checks took 0.66 seconds and used 153.505MB of memory
Psalm was able to infer types for 99.7795% of the codebase
```

This is :fire:

psalm.xml
```diff
@@ -6,55 +6,63 @@
...
 
-        <RedundantCondition errorLevel="suppress" />
+        <PropertyNotSetInConstructor errorLevel="suppress"/>
+        <MissingConstructor errorLevel="error"/>
+        <MissingClosureParamType errorLevel="info"/>
+        <MissingParamType errorLevel="info"/>
 
-        <DocblockTypeContradiction errorLevel="error" />
-        <RedundantConditionGivenDocblockType errorLevel="error" />
+        <RedundantCondition errorLevel="suppress"/>
```
muglug: Is this emitted somewhere erroneously? It's the only issue here that should never be emitted for good code (so if it is, that's a Psalm bug)

Ocramius:
Ok, so my latest changes lead to wild issues in Psalm:

```
ERROR: UndefinedMethod - tests/static-analysis/lazy-loading-value-holder.php:57:25 - Method ProxyManager\Example\VirtualProxy\Foo::sayhello does not exist
    echo $wrappedValue->sayHello();
```

There is no mention of `ProxyManager\Example\VirtualProxy\Foo` in that file, to this feels like a global multi-file shared state issue. Not sure how to reproduce it for psalm to consume.

psalm.xml
Ocramius: Will need to inspect this: I plan to raise the levels before merging

muglug:
> this feels like a global multi-file shared state issue

Yeah something in examples is polluting stuff - I'll find and fix

muglug:
Fixed in https://github.com/vimeo/psalm/commit/0ad57695a893aa1a43a791777ce621aa493164a1

psalm.xml
Ocramius: Bumped all inspection levels

Ocramius:
Hmm, since latest builds, psalm now fails with something like this on 7.4:

```
$ ./vendor/bin/psalm
PHP Fatal error:  Uncaught ParseError: syntax error, unexpected 'Fn' (T_FN), expecting identifier (T_STRING) in /home/travis/build/Ocramius/ProxyManager/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ReturnTypeProvider/ArrayFilterReturnTypeProvider.php:64
```

@muglug I just need to pin this to a stable release with a green build, and then I can release here :+1:

Ocramius:
I'll send a patch to fix the `Fn` issue btw :+1:

dkarlovi:
> a PR fixing itself ? That's inception.

@Ocramius is just that good.

Ocramius:
Green! Now needs a release upstream and we can :ship: \o/

src/ProxyManager/Inflector/ClassNameInflector.php
Ocramius: Keeping this as-is for now.

muglug:
Will hopefully release later today, barring any major impediments

muglug:
Boom: https://github.com/vimeo/psalm/releases/tag/3.4.0

Ocramius:
Thanks! That should be enough to do a release here, if all is green :rocket:

Ocramius:
Relevant CI tasks are GO! :+1:

PR #470, opened 2019-06-21
#469 ensure factory considers `proxyOptions` when generating unique class names

### LINKED ISSUE #469: `ProxyManager\Factory\AbstractBaseFactory#generateProxy()` does not use `$proxyOptions` as part of the class name signature
In practice, the following should be true, but isn't the case right now:

```php
get_class($factory->createProxy(..., ['a' => 'b'])) !== get_class($factory->createProxy(..., ['c' => 'd']))
```

Ocramius:
Fixes #469

### DIFF
--- a/tests/ProxyManagerTest/Factory/AbstractBaseFactoryTest.php
+++ b/tests/ProxyManagerTest/Factory/AbstractBaseFactoryTest.php
@@ -17,6 +17,7 @@ use ProxyManager\Signature\SignatureCheckerInterface;
 use ReflectionMethod;
 use stdClass;
 use Zend\Code\Generator\ClassGenerator;
+use function uniqid;
 
 /**
  * Tests for {@see \ProxyManager\Factory\AbstractBaseFactory}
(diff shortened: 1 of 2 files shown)


Ocramius:
Note: to be cherry-picked into further releases too.

tests/ProxyManagerTest/Factory/AbstractBaseFactoryTest.php
```diff
@@ -169,4 +170,41 @@ public function testGeneratesClass() : void
...
+                self::assertSame($proxyOptions, $parameters['proxyOptions']);
+
+                return true;
+            }))
+            ->will(self::returnValue($generatedClass));
+
+        $this
+            ->proxyAutoloader
+            ->method('__invoke')
+            ->will(self::returnCallback(function (string $className) : bool {
```
malukenho: May it be `static function`?

tests/ProxyManagerTest/Factory/AbstractBaseFactoryTest.php
```diff
@@ -169,4 +170,41 @@ public function testGeneratesClass() : void
...
+                eval('class ' . $className . ' {}');
```
malukenho: Can't this be skipped?

tests/ProxyManagerTest/Factory/AbstractBaseFactoryTest.php
```diff
@@ -169,4 +170,41 @@ public function testGeneratesClass() : void
...
+    {
+        $generateProxy = new ReflectionMethod($this->factory, 'generateProxy');
+
+        $generateProxy->setAccessible(true);
+        $generatedClass = UniqueIdentifierGenerator::getIdentifier('fooBar');
+        $proxyOptions   = [uniqid('key', true) => uniqid('value', true)];
+
+        $this
+            ->classNameInflector
+            ->expects(self::any())
```
malukenho: Why not be more specific here?

tests/ProxyManagerTest/Factory/AbstractBaseFactoryTest.php
Ocramius: Yep

tests/ProxyManagerTest/Factory/AbstractBaseFactoryTest.php
Ocramius: Cannot: a class must be created as part of the test here, or else it will fail to instantiate the proxy

tests/ProxyManagerTest/Factory/AbstractBaseFactoryTest.php
Ocramius: Not part of the scope of this particular test (tested in other tests in the same class IIRC)

Ocramius:
Everything fails, nothing ever works, YAY!

Therefore: merging.

PR #471, opened 2019-06-29
__destruct test PR

odahcam:


Ocramius:
This is the correct way to design the test, but it doesn't seem to reproduce the original issue

tests/language-feature-scripts/lazy-loading-__destruct-throws.phpt
```diff
@@ -0,0 +1,26 @@
...
+    public function __destruct()
+    {
+        # code...
+    }
+}
+
+$factory = new \ProxyManager\Factory\LazyLoadingValueHolderFactory($configuration);
+
+$proxy = $factory->createProxy(Destructable::class, function (& $wrapped, $proxy, $method, array $parameters, & $initializer) {
+    $initializer = null;
```
Ocramius: Maybe this should throw to reproduce the issue?

odahcam: I will have to ask @mnapoli about how container proxies works, because it only happens with this option enabled. 🤔

mnapoli: Hi, I'm not familiar with the issue here sorry. But I don't see why an exception should be thrown, especially if the test case reproduces the issue.

Ocramius: @odahcam the issue is not reproduced here:

```
1) /home/travis/build/Ocramius/ProxyManager/tests/language-feature-scripts/lazy-loading-__destruct-throws.phpt
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
-%SUncaught Error:%sCall to a member function __destruct() on null in %a
/home/travis/build/Ocramius/ProxyManager/tests/language-feature-scripts/lazy-loading-__destruct-throws.phpt:26
```

@mnapoli I'm also having trouble in understanding the full scenario for this issue, sorry...

odahcam: Yesterday I started to think this is a issue with PHP-DI, since the problem only occours when I enable the option `$container->writeProxiesToFiles(true, 'folder/path')`. If @mnapoli could explain what that method does I would understand it better.

@Ocramius maybe here https://github.com/PHP-DI/PHP-DI/issues/669 you can find more information.

At the end I don't know what exactly is happening either, I'm just facing the problem, where one of the proxies generated by PHP-DI using this library throws inside a `__destruct` method because the instance of the class the proxy is for is null inside the proxy class. Yes confusing and strange.

As you can see, things are null here:

image

And then the error occours:

image

Maybe you understand better than me whats happening.

Ocramius: What's the initializer there? It got called, but not reset to `null`, which to me smells like it threw an exception.

odahcam:
```
<?php
$this->initializer5d18ea48bcba2345610290 && $this->initializer5d18ea48bcba2345610290->__invoke($this->valueHolder5d18ea48bcb97124004552, $this, '__destruct', array(), $this->initializer5d18ea48bcba2345610290);
```

image

Is there a way for me to check if it throwed?

Ocramius: I don't store such information, but I'm pretty sure that the engine will aggressively call `__destruct()` anyway.

I'd use a step debugger and put breakpoints on any of the API of the generated code.

Ocramius:
Note: this was closed due to `master` being removed, but the issue stays valid and requires more investigation :-\

odahcam:
Maybe we should open an issue pointing to this PR.

PR #475, opened 2019-07-28
Removed reliance on `Closure::bind(_, null)`

Ocramius:
`Closure::bind()` without a `$this` is deprecated since https://wiki.php.net/rfc/deprecations_php_7_4#unbinding_this_from_non-static_closures,
and therefore we remove reliance on it, or use `static` closures instead.


Ocramius:
Merging here: failure caused by a phpunit and infection interaction, reported at https://github.com/infection/infection/issues/762

PR #483, opened 2019-08-02
Try running PHPUnit like Infection does

sanmai:
As per https://github.com/infection/infection/issues/762

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ before_script:
 
 script:
   - phpdbg -qrr vendor/bin/phpunit --testsuite=unit
-  - ulimit -n 4096 && phpdbg -qrr ./vendor/bin/infection -vvv --test-framework-options='--testsuite=unit' --min-msi=84 --min-covered-msi=86
 
 matrix:
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,46 +5,18 @@ sudo: false
 
 php:
   - 7.4snapshot
-  - nightly
 
 env:
   matrix:
     - DEPENDENCIES="high"
-    - DEPENDENCIES="low"
-  global:
-    - GIT_NAME:  "'Marco Pivetta'"
-    - GIT_EMAIL: ocramius@gmail.com
-    - GH_REF:    github.com/Ocramius/ProxyManager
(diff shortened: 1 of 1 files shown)


Ocramius:
@sanmai no luck? I'm also kinda clueless on what's going on...

sanmai:
OK, let's see if PHPUnit tells us exactly where this is happening.

sanmai:
Interestingly it now gives [a slightly different explanation for the errors](https://travis-ci.org/Ocramius/ProxyManager/builds/566804830#L638-L655). E.g. before it was:

```
1) ProxyManagerTest\GeneratorStrategy\FileWriterGeneratorStrategyTest::testGenerate
PHPUnit\Framework\Exception: Notice: stream_get_contents(): read of 8192 bytes failed with errno=9 Bad file descriptor in Standard input code on line 321
```

And now:

```
1) ProxyManagerTest\GeneratorStrategy\FileWriterGeneratorStrategyTest::testGenerate
PHPUnit\Framework\Exception: [Nothing to execute!]

Caused by
ErrorException: unserialize(): Error at offset 0 of 33 bytes in /home/travis/build/Ocramius/ProxyManager/vendor/phpunit/phpunit/src/Util/PHP/AbstractPhpProcess.php:271
```

PHP version is the same: PHP 7.4.0-dev (cli) (built: Aug  1 2019 09:31:11) ( ZTS )

sanmai:
So far I'm unable to reproduce the issue locally with PHP 7.4.0-dev (cli) (built: Aug  1 2019 23:23:35) ( NTS )

NTS vs. ZTS might have something about it.

sanmai:
But I was trying PHPUnit 8.2.5, and these builds are with 8.3. Starting all over.

sanmai:
All right, now I can reproduce this issue on Docker with php:7.4-rc-cli.

sanmai:
Since we have a reproducible case, I took liberty to report this to the upstream as https://github.com/sebastianbergmann/phpunit/issues/3772. Hope you won't mind.

sanmai:
So you'll have to stay away from 8.3.0 while you use PHPDBG, support for it seems to be on the verge of abandonment.

Ocramius:
Thanks for all this hard work, @sanmai!

Sad for PHPDBG - it started to become more reliable (although not much)

sanmai:
Sounds like the claim to drop support for PHPDBG was withdraw. Thanks for the sound argument for keeping it!

PR #467, opened 2019-06-07
Covered benchmarks via static types through psalm

Ocramius:
TODOs:

 * [x] https://github.com/vimeo/psalm/issues/1812
 * [x] https://github.com/vimeo/psalm/issues/1813 
 * [x] https://github.com/vimeo/psalm/issues/1817
 * [x] https://github.com/vimeo/psalm/issues/1818
 * [x] pin to stable `vimeo/psalm`
 * [x] remove as many `/** @var */` and `/** @psalm-var */` inline declarations as possible

PR #484, opened 2019-08-06
Fix deprecation PHP 7.4

nicolas-grekas:
This PR fixes a deprecation notice that is triggered when the package is used on PHP 7.4.
I'd prefer having this patch on 2.1, but it's not maintained anymore, right? Looks like 2.2 is?
Patch borrowed from #467 so this should reduce merge/cherry-pick conflicts.
Thanks for considering!

### DIFF
--- a/src/ProxyManager/Generator/Util/ProxiedMethodReturnExpression.php
+++ b/src/ProxyManager/Generator/Util/ProxiedMethodReturnExpression.php
@@ -16,7 +16,15 @@ final class ProxiedMethodReturnExpression
 {
     public static function generate(string $returnedValueExpression, ?\ReflectionMethod $originalMethod) : string
     {
-        if ($originalMethod && 'void' === (string) $originalMethod->getReturnType()) {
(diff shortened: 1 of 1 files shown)


Ocramius:
I already got these fixed in `master` for `2.5.0`. Will likely cut a release from there during the weekend.

Ocramius:
Ref: https://github.com/Ocramius/ProxyManager/pull/475
Ref: https://github.com/Ocramius/ProxyManager/pull/467
Ref: https://github.com/Ocramius/ProxyManager/pull/465

nicolas-grekas:
Hum, I see. This one is likely not going to be accepted, isn't it?

Ocramius:
No, I think people can safely upgrade to `2.5.0`, which is fully BC compliant, once I cut it.

I got only a test failure due to a PHPUnit regression, so I held it off, but I'll see if I can fix it while at Froscon.

nicolas-grekas:
OK, thanks for the feedback, closing then. Happy Froscon :)

PR #486, opened 2019-08-10
Remove process isolation from tests around file writer generators

Ocramius:
See https://github.com/sebastianbergmann/phpunit/issues/3772

~~As discussed with @sebastianbergmann, PHPUnit 8.3.4 may fix this issue, but until then,
we lock onto an older version to have green builds to validate a release~~

For now, we simply remove process isolation to simplify the issue overall

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     },
     "require-dev": {
         "ext-phar":                     "*",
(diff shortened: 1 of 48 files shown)


Ocramius:
Hmm, doesn't seem to be the culprit...

Ocramius:
GREEN! FINALLY!

Issue #488, opened 2019-08-24
Drop compiled documentation


Ocramius:
As discussed briefly with @malukenho, the compiled docs (in the `gh-pages` branch) can probably be dropped in favour of just letting folks browse the `.md` files (which render quite nicely).

Given the amount of added dependencies coming in through the build tools used for publishing these pages, I think @malukenho's suggestion is sound.

Having a webpage is quite redundant right now: maybe we can just set up a redirect on 404, pointing to the repository `docs/` dir?

Ocramius:
@malukenho assigned to you just to have your opinion.

malukenho:
I fully agree on that, I will pick this one up.

Ocramius:
@malukenho no need for it: I just needed your opinion 👍

Ocramius:
Handled in #490 and #491

PR #490, opened 2019-08-28
Just redirect right away from the "index" doc

malukenho:


malukenho:
Btw it looks like the couscous deploy is not working.
And we should remove the HTML docs link reference form the repository description.

Ocramius:
> Btw it looks like the couscous deploy is not working.

I'd suggest dropping couscous completely, and having custom HTML deployed to `gh-pages`

malukenho:
@Ocramius can you change that?
Screen Shot 2019-08-28 at 12 44 57 PM

Ocramius:
Most certainly. Just pointing it to `docs`?

malukenho:
I think you can remove the link completely. I've created #491 with HTML redirects

malukenho:
Linking #488

PR #502, opened 2019-09-14
Documented basic immutable/mutation-free interfaces and services in the package

Ocramius:
TODO:

 * [x] ~~Mark as much as possible as immutable~~ (good enough for now)
 * [x] ~~Write down list of architectural limitations due to upstream inherited mutability~~ (not needed in this patch)
 * [x] https://github.com/vimeo/psalm/issues/2138


malukenho:
That looks awesome already

Ocramius:
Shipping this as-is: will add more immutability/purity markers later, if possible.

PR #522, opened 2019-11-15
Feature: integration with webimpress/safe-writer

michalbundyra:
Use external library to write files

### DIFF
--- a/src/ProxyManager/Exception/FileNotWritableException.php
+++ b/src/ProxyManager/Exception/FileNotWritableException.php
@@ -34,6 +34,6 @@ class FileNotWritableException extends UnexpectedValueException implements Excep
 
     public static function exception(FileWriterException $e) : self
     {
-        return new self($e->getMessage(), $e->getCode(), $e);
+        return new self($e->getMessage(), (int) $e->getCode(), $e);
     }
 }
--- a/src/ProxyManager/Exception/FileNotWritableException.php
+++ b/src/ProxyManager/Exception/FileNotWritableException.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace ProxyManager\Exception;
 
 use UnexpectedValueException;
+use Webimpress\SafeWriter\Exception\ExceptionInterface as FileWriterException;
 use function sprintf;
 
 /**
@@ -29,5 +30,10 @@ class FileNotWritableException extends UnexpectedValueException implements Excep
             . 'either the directory does not exist, or it is not writable',
             $directory
         ));
+    }
+
+    public static function exception(FileWriterException $e) : self
+    {
+        return new self($e->getMessage(), $e->getCode(), $e);
     }
 }
--- a/src/ProxyManager/Exception/FileNotWritableException.php
+++ b/src/ProxyManager/Exception/FileNotWritableException.php
@@ -13,6 +13,9 @@ use function sprintf;
  */
 class FileNotWritableException extends UnexpectedValueException implements ExceptionInterface
 {
+    /**
+     * @deprecated
+     */
     public static function fromInvalidMoveOperation(string $fromPath, string $toPath) : self
     {
         return new self(sprintf(
@@ -23,6 +26,9 @@ class FileNotWritableException extends UnexpectedValueException implements Excep
         ));
     }
 
+    /**
+     * @deprecated
+     */
     public static function fromNotWritableDirectory(string $directory) : self
     {
         return new self(sprintf(
--- a/src/ProxyManager/Exception/FileNotWritableException.php
+++ b/src/ProxyManager/Exception/FileNotWritableException.php
@@ -38,8 +38,8 @@ class FileNotWritableException extends UnexpectedValueException implements Excep
         ));
     }
 
-    public static function exception(FileWriterException $e) : self
+    public static function fromPrevious(FileWriterException $previous) : self
     {
-        return new self($e->getMessage(), (int) $e->getCode(), $e);
+        return new self($previous->getMessage(), 0, $previous);
     }
 }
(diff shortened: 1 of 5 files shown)


michalbundyra:
@Ocramius I've noted these, and I've decided to create v2.0.0 with PHP 7.1+ support, both channels will be maintained.
The change in v2.0 are gonna be:
- typehints/rth
- strict type declaration
- ExceptionInterface extends Throwable <-- this is gonna solve that issue you see above.

Will ping you one it's updated. Than

Ocramius:
Just noticed it, but `EvaluatingGeneratorStrategy` would need the same kind of patch

michalbundyra:
@Ocramius ok, will update it today, just about release v2.0.0 :)

Ocramius:
Easy easy, we're not in a hurry, it was just about something that I noticed to be missing from this patch :+1:

michalbundyra:
@Ocramius I've updated the PR but still we are getting one error:

```
ERROR: PossiblyInvalidArgument - src/ProxyManager/Exception/FileNotWritableException.php:37:43 - Argument 2 of ProxyManager\Exception\FileNotWritableException::__construct expects int, possibly different type int|string provided
830        return new self($e->getMessage(), $e->getCode(), $e);
```

I believe it is invalid as per documentation - getCode return `int` (not `int|string`):

https://www.php.net/manual/en/throwable.getcode.php

michalbundyra:
Ok, I see... `Exception::getCode() : mixed`
https://www.php.net/manual/en/exception.getcode.php

src/ProxyManager/Exception/FileNotWritableException.php
```diff
@@ -30,4 +31,9 @@ public static function fromNotWritableDirectory(string $directory) : self
             $directory
         ));
     }
+
+    public static function exception(FileWriterException $e) : self
```
Ocramius: I would probably call this `fromPrevious` or such.

If the other named constructors are no longer in use, they should be `@deprecated`.

Let's also rename `$e` to `$previous`

src/ProxyManager/Exception/FileNotWritableException.php
```diff
@@ -30,4 +31,9 @@ public static function fromNotWritableDirectory(string $directory) : self
...
+    {
+        return new self($e->getMessage(), (int) $e->getCode(), $e);
```
Ocramius: The error code can be safely skipped here: passing it on to the new exception doesn't convey anything useful downstream

src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php
```diff
@@ -43,11 +40,9 @@ public function generate(ClassGenerator $classGenerator) : string
 
         // @codeCoverageIgnoreStart
         if (! $this->canEval) {
-            $fileName = tempnam(sys_get_temp_dir(), 'EvaluatingGeneratorStrategy.php.tmp.');
+            $fileName = __DIR__ . DIRECTORY_SEPARATOR . 'EvaluatingGeneratorStrategy.php.tmp';
```
Ocramius: Let's avoid relying on `DIRECTORY_SEPARATOR`: a `/` is more consistent and readable too

michalbundyra: Done now - after your approval

Ocramius:
Sorry to bother further: can you drop tests for the deprecated API in `FileNotWritableExceptionTest`? :S

michalbundyra:
@Ocramius done. Added also test for new method in FileNotWritableException, but - because of mock - we are getting error there (psalm). How I can fix it?

Somehow it was not working with prophecy mocking:
```php
        $previous = $this->prophesize(FileWriterException::class)
            ->willBeConstructedWith(['Previous exception message'])
            ->reveal();
```

it might be a bug in prophecy, but haven't had time to investigate more.

Ocramius:
Prophecy mocking should be avoided: it's too magic and misses psalm annotations required for type inference.

Traditional PHPUnit mocks would be much simpler/better ;-)

Ocramius:
```
        $previous = $this->prophesize(FileWriterException::class)
            ->willBeConstructedWith(['Previous exception message'])
            ->reveal();
```

Probably no need to mock the exception itself, and expect the ctor call? I'd rather instantiate the real thing here.

michalbundyra:
I've used traditional PHPUnit mock but we are getting psalm error anyway... See build logs.

tests/ProxyManagerTest/Exception/FileNotWritableExceptionTest.php
```diff
@@ -15,25 +16,16 @@
  */
 final class FileNotWritableExceptionTest extends TestCase
 {
-    public function testFromInvalidMoveOperation() : void
+    public function testFromPrevious() : void
     {
-        $exception = FileNotWritableException::fromInvalidMoveOperation('/tmp/a', '/tmp/b');
+        $previous = $this->getMockBuilder(FileWriterException::class)
```
Ocramius: I'd probably instantiate a real `FileWriterException` here, instead of a mock. Psalm chokes here because of the fluent interface usage, but getting rid of the mock directly would simplify things enough.

michalbundyra: There are 4 different exceptions which implements the same interface:
https://github.com/webimpress/safe-writer/tree/master/src/Exception
All these exceptions have private constructor, and all public static methods are marked as internal.
I believe mocking in that place is the most appropriate, and we should skip somehow psalm error here.

michalbundyra: Dropped fluent interface usage and it works now ! :) Thanks, @Ocramius!

Ocramius: Nice! BTW, since we're mocking only the interface, the `final public function getMessage()` isn't really enforced anyway (in theory)

Ocramius:
Thanks @michalbundyra!

PR #537, opened 2019-12-11
Use ReflectionClass to create proxied class instance inside PublicScopeSimulator

### LINKED ISSUE #536: Accessing properties of Serializable proxy
When accessing a property of a proxy class that implements `Serializable`, the PHP error `Erroneous data format for unserializing 'class'` is thrown.

`PublicScopeSimulator::getTargetObject()` uses `'unserialize(sprintf(\'O:%d:"%s":0:{}\', strlen(get_parent_class($this)), get_parent_class($this)))'` to create class instance and is related to this PHP bug report https://bugs.php.net/bug.php?id=68099 which is marked as 'Wont Fix'.

As explained in bug there is different format used for basic object serialize/unserialize and the one that uses `Serializable` interface. I guess proper solution would be to use [ReflectionClass::newInstanceWithoutConstructor](https://www.php.net/manual/en/reflectionclass.newinstancewithoutconstructor.php) to create new empty instance of an object, unless there is reason why `unserialize` is needed.

Ludo444:
closes #536

### DIFF
--- a/src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
+++ b/src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
@@ -112,7 +112,7 @@ class PublicScopeSimulator
             return '$this->' . $valueHolder->getName();
         }
 
-        return 'unserialize(sprintf(\'O:%d:"%s":0:{}\', strlen(get_parent_class($this)), get_parent_class($this)))';
+        return '$realInstanceReflection->newInstanceWithoutConstructor()';
     }
 
     /**
--- a/tests/ProxyManagerTest/ProxyGenerator/Util/PublicScopeSimulatorTest.php
+++ b/tests/ProxyManagerTest/ProxyGenerator/Util/PublicScopeSimulatorTest.php
@@ -40,7 +40,7 @@ if (! $realInstanceReflection->hasProperty($foo)) {
     return;
 }
 
-$targetObject = unserialize(sprintf('O:%d:"%s":0:{}', strlen(get_parent_class($this)), get_parent_class($this)));
+$targetObject = $realInstanceReflection->newInstanceWithoutConstructor();
 $accessor = function & () use ($targetObject, $name) {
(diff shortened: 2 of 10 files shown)


Ludo444:
Yes, error I get is the php bug in linked issue - `Warning: Erroneous data format for unserializing`.

I will look into those feature tests later, hopefully some time this week since not really familiar with it.

Ocramius:
> I will look into those feature tests later, hopefully some time this week since not really familiar with it.

They are relatively simple scripts: the `EXPECT` part contains expected output (or output regex): I can help with that. PHPUnit runs them in isolation if you pass in `./vendor/bin/phpunit tests/language-feature-scripts`

Ludo444:
Something like this is fine?

```
// tests/language-feature-scripts/access-interceptor-scope-localizer-serialized-class-private-property-isset.phpt

--TEST--
Verifies that generated access interceptors doesn't throw PHP Warning on Serialized class private property direct isset check
--FILE--
<?php

require_once __DIR__ . '/init.php';

class Kitchen implements \Serializable
{
    private $sweets = 'candy';

    function serialize() 
    {
        return $this->sweets;
    }

    function unserialize($serialized)
    {
        $this->sweets = $serialized;
    }
}

$factory = new \ProxyManager\Factory\AccessInterceptorScopeLocalizerFactory($configuration);

$proxy = $factory->createProxy(new Kitchen());

var_dump(isset($proxy->sweets));
?>
--EXPECT--
bool(false)
```

Ocramius:
Yes, something like that looks fine. If I understand it correctly, that crashes with current `master` state.

Ludo444:
Yes, under current `master` it fails, and I suppose the added test should be passing in new version

Ocramius:
Yep: smash it in there, and I'll get it into a `3.5.1` or such :+1:

Ocramius:
Thanks @Ludo444! I've just tagged 2.5.1 with your changes in it \o/ 

https://github.com/Ocramius/ProxyManager/releases/tag/2.5.1

PR #546, opened 2020-01-01
Migrating zendframework to laminas

orklah:
As discussed on #542, using zendframework is now discouraged in favor of laminas.

I kept zendframework in the conflict section but I'm not an expert on this so I may be wrong


Ocramius:
This still requires changing all `Zend\\` references to `Laminas\\` references (should be a find-replace task)

orklah:
I replaced all usages and mentions in code and docs. There is just one left in the changelog on purpose.

Ocramius:
LGTM! Well done, thanks! :+1:

examples/remote-proxy.php
```diff
@@ -6,13 +6,13 @@
...
+use Laminas\XmlRpc\Client;
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
-if (! class_exists('Zend\XmlRpc\Client')) {
-    echo "This example needs Zend\\XmlRpc\\Client to run. \n In order to install it, "
+if (! class_exists('Laminas\XmlRpc\Client')) {
+    echo "This example needs Laminas\\XmlRpc\\Client to run. \n In order to install it, "
         . "please run following:\n\n"
         . "\$ php composer.phar require zendframework/zend-xmlrpc:2.*\n\n";
```
victorlap: Should this line be changed as well?

orklah: The Class can't exists if it's not imported and we can't import it without having it in composer so I believe it should.

However, the require command still mentions zendframework. This should be changed too

Ocramius: Handled in #550

PR #548, opened 2020-01-02
Fixed code style after replacing Zend with Laminas

alfredbez:
We have a bunch of code style violations after #546, because the `use` statements are not sorted correctly anymore. I've just executed `vendor/bin/phpcbf` to fix those errors.

Maybe it makes sense to revert 4c66d45a88538dcb96814b046386de8b49373e54, since it works now again?

### DIFF
--- a/tests/ProxyManagerTest/Functional/AccessInterceptorScopeLocalizerFunctionalTest.php
+++ b/tests/ProxyManagerTest/Functional/AccessInterceptorScopeLocalizerFunctionalTest.php
@@ -226,8 +226,7 @@ final class AccessInterceptorScopeLocalizerFunctionalTest extends TestCase
     /**
      * @dataProvider getPropertyAccessProxies
      */
-    public function testPropertyWriteAccess(object $instance, AccessInterceptorInterface $proxy, string $publicProperty
-    ) : void
+    public function testPropertyWriteAccess(object $instance, AccessInterceptorInterface $proxy, string $publicProperty) : void
     {
         $newValue               = uniqid('value', true);
         $proxy->$publicProperty = $newValue;
@@ -239,8 +238,7 @@ final class AccessInterceptorScopeLocalizerFunctionalTest extends TestCase
     /**
      * @dataProvider getPropertyAccessProxies
      */
-    public function testPropertyExistence(object $instance, AccessInterceptorInterface $proxy, string $publicProperty
-    ) : void
+    public function testPropertyExistence(object $instance, AccessInterceptorInterface $proxy, string $publicProperty) : void
     {
         self::assertSame(isset($instance->$publicProperty), isset($proxy->$publicProperty));
         $this->assertProxySynchronized($instance, $proxy);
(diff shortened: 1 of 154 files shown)


Ocramius:
CS is not critical: as you see in the CI pipeline, I use `|| true` there, since I value more the type checking provided by psalm.

tests/ProxyManagerTest/Functional/AccessInterceptorScopeLocalizerFunctionalTest.php
```diff
@@ -523,7 +520,7 @@ public function testWillInterceptAndReturnEarlyOnVoidMethod() : void
                         VoidCounter $instance,
                         string $method,
                         array $params,
-                        ?bool & $returnEarly
+                        ?bool &$returnEarly
```
Ocramius: These should be reverted: `& $returnEarly` is the correct formatting when using by-ref params (which are otherwise much more elusive to the eye)

Ocramius: Note: the CS rule changing this can be disabled, or maybe we can upgrade to a newer `doctrine/coding-standard` version to see if that already fixes it.

tests/ProxyManagerTest/Functional/AccessInterceptorScopeLocalizerFunctionalTest.php
```diff
@@ -312,7 +309,7 @@ public function testWillModifyByRefRetrievedPublicProperties() : void
         $instance = new ClassWithPublicProperties();
         $proxy    = (new AccessInterceptorScopeLocalizerFactory())->createProxy($instance);
 
-        $variable = &$proxy->property0;
+        $variable = & $proxy->property0;
```
alfredbez: not sure if this is ok

Ocramius: Seems OK

tests/ProxyManagerTest/GeneratorStrategy/FileWriterGeneratorStrategyTest.php
```diff
@@ -47,7 +47,7 @@ protected function setUp() : void
         parent::setUp();
 
         $this->tempDir              = tempnam(sys_get_temp_dir(), 'FileWriterGeneratorStrategyTest');
-        $this->originalErrorHandler = static function () : bool {
+        $this->originalErrorHandler = static function () : void {
```
Ocramius: This is supposed to stay `: bool`

alfredbez: I've reverted that, but just for my understanding: This is needed because you defined the return type of the second argument using `@psalm-param` here: https://github.com/Ocramius/ProxyManager/blob/dfc740709aab9a334b75ba09ef948830c3be2aa9/src/ProxyManager/Factory/LazyLoadingGhostFactory.php#L87

Is that right? The static function doesn't really return anything or is there some magic I can't see?

Ocramius: Correct: the function doesn't return anything, but a `callable() : bool` is expected by callees

PR #558, opened 2020-01-21
Added stream wrapper support to FileLocator

Bittarman:
This pr allows the FileLocator to support stream wrappers.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,8 @@
         "phpbench/phpbench":            "^0.16.10",
         "infection/infection":          "^0.15.0",
         "symfony/console":              "^4.4.2",
-        "vimeo/psalm":                  "3.7.0"
+        "vimeo/psalm":                  "3.7.0",
+        "mikey179/vfsstream": "^1.6.8"
     },
     "suggest": {
         "ocramius/generated-hydrator": "To have very fast object to array to object conversion for ghost objects",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -22,7 +22,8 @@
         "php":                       "7.4.*",
         "laminas/laminas-code":      "^3.4.1",
         "ocramius/package-versions": "^1.5.1",
-        "webimpress/safe-writer":    "^2.0"
+        "webimpress/safe-writer":    "^2.0",
+        "mikey179/vfsstream": "^1.6"
```
Ocramius: Should be a Dev dependency

src/ProxyManager/FileLocator/FileLocator.php
```diff
@@ -24,10 +25,12 @@ public function __construct(string $proxiesDirectory)
         $absolutePath = realpath($proxiesDirectory);
 
         if ($absolutePath === false) {
-            throw InvalidProxyDirectoryException::proxyDirectoryNotFound($proxiesDirectory);
+            if (! is_dir($proxiesDirectory)) {
```
Ocramius: Hmm, this is an added check: wasn't the problem the realpath call?

Bittarman: Yes, but as you mentioned you also needed it to deal with symlinks, this seemed to be the better option, and would result in less change of behaviour.
is_dir will work with a stream wrapper, while realpath will not.

composer.json
Bittarman: oops, sorry. fixed.

src/ProxyManager/FileLocator/FileLocator.php
Ocramius: Discussed in chat: I didn't see (during review) that `$proxiesDirectory` is no longer discarded right away

Ocramius:
Thanks @Bittarman!

PR #567, opened 2020-02-03
Better type inference for used `class-string`s

Ocramius:


composer.json
```diff
@@ -39,7 +39,7 @@
         "phpbench/phpbench":            "^0.16.10",
         "infection/infection":          "^0.15.0",
         "symfony/console":              "^4.4.2",
-        "vimeo/psalm":                  "3.7.0"
+        "vimeo/psalm":                  "dev-master"
```
Ocramius: We'll need to pin this before merging

PR #573, opened 2020-03-09
WIP: Added test stubs for verifying typed read-only properties behavior

Ocramius:
Ref: https://wiki.php.net/rfc/write_once_properties
Ref: https://github.com/php/php-src/pull/5186


kocsismate:
@Ocramius Thank you very much! I'll try this patch with my patch XD

Ocramius:
This RFC was declined, so for now I'm going to close this.

This would've been so much more relevant than most other RFCs affecting PHP 8.0 :-(

PR #577, opened 2020-03-26
Fix signatures to static

muglug:
Latest master version of Psalm is a bit stricter about when you can use `static`.


Ocramius:
Thanks!

PR #578, opened 2020-03-26
fix gitattributes due to changed files

Tobion:


.gitattributes
```diff
@@ -1,5 +1,6 @@
+/.github export-ignore
+/docs export-ignore
```
Ocramius: These should still be exported, IMO

Tobion: sure fine for me. it wasn't exported once before due to /html-docs export-ignore. i guess the /examples should be removed from the gitattributes as well?

Ocramius: Yes please! HTML docs are long gone. Thanks for looking into this!

Tobion: done

PR #583, opened 2020-04-05
Fix newly-prevented issues

muglug:
Psalm had a bit of an issue understanding interpreting partially matching templates (`$a` was previously inferred as `object` in [this example](https://psalm.dev/r/1ad54ea020), now it's inferred as `A`). 

This means that Psalm (latest master) now finds a few additional issues in the tests and example code.

### DIFF
--- a/tests/ProxyManagerBench/LazyLoadingGhostPropertyAccessBench.php
+++ b/tests/ProxyManagerBench/LazyLoadingGhostPropertyAccessBench.php
@@ -56,6 +56,9 @@ final class LazyLoadingGhostPropertyAccessBench
     /** @var ClassWithMixedProperties&LazyLoadingInterface */
     private ClassWithMixedProperties $initializedMixedPropertiesProxy;
 
+    /**
+     * @psalm-suppress RedundantConditionGivenDocblockType
+     */
     public function setUp() : void
     {
         $emptyClassProxy                     = $this->buildProxy(EmptyClass::class);
--- a/tests/ProxyManagerBench/LazyLoadingValueHolderPropertyAccessBench.php
+++ b/tests/ProxyManagerBench/LazyLoadingValueHolderPropertyAccessBench.php
@@ -44,6 +44,9 @@ final class LazyLoadingValueHolderPropertyAccessBench
     /** @var ClassWithMixedProperties&VirtualProxyInterface */
     private ClassWithMixedProperties $initializedMixedPropertiesProxy;
 
+    /**
+     * @psalm-suppress RedundantConditionGivenDocblockType
+     */
     public function setUp() : void
     {
         $emptyClassProxy                  = $this->buildProxy(EmptyClass::class);
--- a/tests/ProxyManagerTest/Functional/LazyLoadingValueHolderFunctionalTest.php
+++ b/tests/ProxyManagerTest/Functional/LazyLoadingValueHolderFunctionalTest.php
@@ -280,7 +280,7 @@ final class LazyLoadingValueHolderFunctionalTest extends TestCase
 
         $proxy = (new LazyLoadingValueHolderFactory())->createProxy(
             BaseClass::class,
-            static function (?object & $wrappedInstance) use (& $counter) : bool {
+            static function (?BaseClass & $wrappedInstance) use (& $counter) : bool {
                 $wrappedInstance = new BaseClass();
 
                 /** @var int $counter */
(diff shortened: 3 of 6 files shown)


muglug:
I guess this will have to be delayed until the next version of Psalm is released.

tests/ProxyManagerBench/LazyLoadingValueHolderPropertyAccessBench.php
```diff
@@ -44,6 +44,9 @@ final class LazyLoadingValueHolderPropertyAccessBench
     /** @var ClassWithMixedProperties&VirtualProxyInterface */
     private ClassWithMixedProperties $initializedMixedPropertiesProxy;
 
+    /**
+     * @psalm-suppress RedundantConditionGivenDocblockType
```
Ocramius, line 48: I think the assertions can be dropped here - will investigate

tests/ProxyManagerBench/LazyLoadingGhostPropertyAccessBench.php
```diff
@@ -56,6 +56,9 @@ final class LazyLoadingGhostPropertyAccessBench
     /** @var ClassWithMixedProperties&LazyLoadingInterface */
     private ClassWithMixedProperties $initializedMixedPropertiesProxy;
 
+    /**
+     * @psalm-suppress RedundantConditionGivenDocblockType
```
Ocramius, line 60: I think the assertions can be dropped here - will investigate

tests/ProxyManagerTest/Functional/LazyLoadingValueHolderFunctionalTest.php
```diff
@@ -280,7 +280,7 @@ public function testWillAllowMultipleProxyInitialization() : void
 
         $proxy = (new LazyLoadingValueHolderFactory())->createProxy(
             BaseClass::class,
-            static function (?object & $wrappedInstance) use (& $counter) : bool {
+            static function (?BaseClass & $wrappedInstance) use (& $counter) : bool {
```
Ocramius, line 283: This should still be valid with `?object`.

In fact, in libraries like `doctrine/orm`, this will work with numerous types of objects

tests/ProxyManagerTest/Functional/LazyLoadingValueHolderFunctionalTest.php
```diff
@@ -375,15 +375,19 @@ static function (?object & $wrappedInstance) : bool {
...
 
     /**
+     * @template TClass
+     *
+     * @psalm-param (CallableInterface&Mock)|null $initializerMatcher
+     * @psalm-param class-string<TClass> $className
+     * @psalm-param TClass $realInstance
+     *
      * @return Closure(
-     *  object|null,
```
Ocramius, line 379: `object|null` should be accepted here: the closure is indeed designed to be more permissive

tests/ProxyManagerTest/Functional/LazyLoadingValueHolderFunctionalTest.php
```diff
@@ -402,7 +406,11 @@ private function createInitializer(string $className, object $realInstance, ?Moc
                 );
         }
 
-        return static function (
+        return
+        /**
+         * @psalm-param ?TClass $wrappedObject
```
Ocramius, line 411: Similar to above, closure parameter should accept any object here.

The factory states:

```
     * @psalm-param Closure(
     *   RealObjectType|null=,
     *   RealObjectType&ValueHolderInterface<RealObjectType>&VirtualProxyInterface=,
     *   string=,
     *   array<string, mixed>=,
     *   ?Closure=
     * ) : bool $initializer
```

In theory, `function (?object $object) : bool {...` should be compatible, no?

tests/static-analysis/lazy-loading-value-holder.php
```diff
@@ -20,7 +20,7 @@ public function sayHello() : string
         ->createProxy(
             MyProxiedClass::class,
             static function (
-                ?object & $instance,
+                ?MyProxiedClass & $instance,
```
Ocramius, line 23: Similar to above: `?object` should be acceptable

tests/ProxyManagerTest/Functional/LazyLoadingValueHolderFunctionalTest.php
muglug: Yeah, I think these should actually both have the same error: https://psalm.dev/r/1785fafb65

PR #600, opened 2020-05-12
Fix documentation code samples

malukenho:
I've used https://github.com/codelicia/xulieta.
I'm not sure about the usage of the tool for this project but could be an idea.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -176,7 +176,7 @@ By-ref variadic arguments are also supported:
 ```php
 class SayHello
 {
-    public function hello(string ... & $names)
+    public function hello(string &...$names)
     {
         foreach ($names as & $name) {
             $name = 'hello, ' . $name;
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -208,7 +208,7 @@ $proxyClass = $proxyGenerator->generateProxy(
     new ClassGenerator('ProxyClassName')
 );
 
-eval('<?php ' . $proxyClass->generate());
+eval('<? ' . $proxyClass->generate());
```
Ocramius, line 211: This should really not use short tags

malukenho: Yep. https://github.com/codelicia/xulieta/issues/19 I can open the `php` tag at the beginning of the code sample as a workaround, or just ignore this one as we don't have the tool here.

malukenho: Which one do you prefer?

Ocramius: I think the tool has a bug here: this is a string inside an `eval()` block, but xulieta caught it as a code delimiter

malukenho: Fixed by opening the php tag at the beginning of the code sample. I do have an issue open for that.

PR #606, opened 2020-05-15
Run xulieta in CI

malukenho:


.github/workflows/tests.yml
```diff
@@ -40,6 +40,9 @@ jobs:
       - name: Run PhpCs
         run: ./vendor/bin/phpcs
 
+      - name: Run Xulieta
+        run: ./vendor/bin/xulieta check:erromeu .
```
Ocramius, line 44: erromeu?!

malukenho: https://github.com/codelicia/xulieta/blob/master/meme.jpg

Ocramius: i don't mind, but meh...

PR #623, opened 2020-08-26
Fix Versions checking for ocramius/package-versions instead of ocramius/proxy-manager

nicolas-grekas:
Unless I missed something, this looks like a typo.
A local project is failing with `Package "ocramius/package-versions" is not installed` currently.


Ocramius:
Most certainly a bug (too many libs), thanks for noticing!

PR #631, opened 2020-11-13
Delete unnecessary (?) Version class

Seldaek:
Seeing the latest round of 🤬 is most likely in large part due to this package requiring composer 2, it made me think.. Why is it?

The way I see it, there is no direct benefit of a package exposing it's own version via `getVersion`, now that Composer can do it.

If you remove this here, then people using Composer 1 can just do that, and if they need to know the version they'd need to use Composer 2, or install package-versions, whatever it's up to them. People using Composer 2 can use InstalledVersions themselves to fetch the package's version if they are interested in it for whatever reason. I can see that doing proper version checks is better than feature-detection with classes or methods if you need to support multiple versions of this library.

I see that proxyManagerVersion is also used to ensure proxies are regenerated when the version of this lib changes. I guess this might make the change unacceptable to you. It'd require maintaining a const ABI_VERSION of sorts that changes whenever proxy generation code changes which I guess is on almost every commit here so probably a pain to keep updating. Feel free to close this, I am opening it mostly to discuss the concept.

TL;DR: It's less code here, and moves the responsibility for version fetching where it belongs IMO.


nicolas-grekas:
:100: here

see #643 for a possible implementation with a BC layer.

Ocramius:
This is exactly where I was before `Version` was introduced.

`ProxyManager\Version` was introduced in `1.0.0-beta`:

https://github.com/Ocramius/ProxyManager/blob/3608247f2dc8ebbc3d22e6289745b1ddc7f1059a/src/ProxyManager/Version.php#L29-L31

That quickly became one more roadblock in maintaining the library, at which point I created `ocramius/package-versions` for the sole purpose of getting rid of these use-cases:

https://github.com/Ocramius/ProxyManager/commit/9af84d6d4ec954ca57702633c3b98ee476f42e83#diff-163bc3b2c01cab664cfc3ed293afb308ae99b8d8b81d05c65734c154c3186c9c (#285)
https://github.com/Ocramius/ProxyManager/blob/9af84d6d4ec954ca57702633c3b98ee476f42e83/src/ProxyManager/Version.php#L33-L54

That finally got me rid of the maintained constant:

https://github.com/Ocramius/ProxyManager/commit/8b80b054a870805add4c0fe4e6128040a0d7a408#diff-163bc3b2c01cab664cfc3ed293afb308ae99b8d8b81d05c65734c154c3186c9c

I've then moved to `composer-runtime-api:^2`:

https://github.com/Ocramius/ProxyManager/blob/d0cc213b3c2b5482ed47d3910be7552ebc5de5fe/src/ProxyManager/Version.php#L35-L39

Then @nicolas-grekas has fixed an obvious SNAFU in 1b972ab5a8f997088ba57e69d48053b657f75578 (#623).

From there on, it has been a non-problem, and still isn't a problem, since `ocramius/package-versions` works as expected, unless there are bugs I'm unaware of.

Closing as `invalid`, since the package is operating as-expected.

PR #640, opened 2020-12-18
Upgrade to PHPUnit ^9.4

martinssipenko:


composer.lock
```diff
@@ -5815,37 +5981,39 @@
                     "type": "tidelift"
                 }
             ],
-            "time": "2020-07-05T13:08:05+00:00"
+            "time": "2020-10-27T19:25:29+00:00"
         },
         {
             "name": "vimeo/psalm",
-            "version": "3.12.2",
+            "version": "3.18.2",
```
Ocramius: Please do a `git revert` on `composer.lock`, then try a `composer update phpunit/phpunit` to isolate out this `vimeo/psalm` upgrade.

`vimeo/psalm` is correct here (in the issues reported in this patch), but that's my problem to solve later :-)

Ocramius (changes requested):

martinssipenko:
Was not as easy as you described because of PHPUnit's and Psalms' common dependencies, but I think I managed to achieve to upgrade PHPUnit and only its dependencies that had to be upgraded.

Ocramius (approved):
LGTM, thanks @martinssipenko!

Issue #630, opened 2020-11-13
I am willing to help maintain older branches of this package


nicolas-grekas:
The versioning policy of this package creates a lot of friction with other OSS projects.

The option to "pay for maintenance" that is advertised in the readme doesn't work for the other OSS maintainers: we do spend too much time over our weekends working around this policy. But we can "pay you" with our time.

One reason for the policy that I understand is that you don't have the resources to maintain eg a workable 2.1 branch, that would support a wider range of PHP versions, from 7.1 to upper.

The code in this repo is great and I certainly do not want to duplicate the effort.

I see only two solutions to the current situation:
1. maintain a fork that I would spend part of my OSS time syncing with this one. I'd backport patches to make them work on PHP >=7.1. [I did it](https://github.com/Ocramius/ProxyManager/compare/2.11.x...nicolas-grekas:php71) on my last weekend and that was pretty easy, so this looks sustainable on my side. Many Linux distributions do a similar thing already, maintaining their fork in a way that fits their policies.
2. become a co-maintainer of this repo, especially responsible for supporting a wider range of PHP+Composer versions. Thus this issue.

The third option, which is accepting the existing policy, is too painful. See also [similar discussions around the package-versions](https://github.com/Ocramius/PackageVersions/pull/133) package. I wish there was more cooperation between OSS maintainers. That also means between you and me.

As I posted elsewhere, I admire your work at many levels. I'm just sooooo annoyed by the policy (and I know I'm not the only one) that I want to do something about it. Please help us.

mpalourdio:
Not trying to be rude, but why making things work with nearly dead PHP versions or soon deprecated libraries should be done by Symfony OSS maintainers over their weekend ? If this library is such centric for the Symfony ecosystem, Symfony could support paid work on this library for backward compatibility to suit its own dependencies constraints.

That's why companies pay for RedHat support, and geeks play with Arch, because as developers, we love bleeding edge technologies :)

asgrim:
@nicolas-grekas FWIW, I really appreciate this as a much better and more cooperative way of reaching out for discussion, kudos :+1: so much better!

bwoebi:
It is a bit sad that the new holy war of the PHP ecosystem is now system version requirements.

I _do_ agree that in an ideal world, we are always everywhere on the newest version.

[...]

Issue #468, opened 2019-06-14
[From PHP-DI] `__destruct` magic method being called before lazy Proxy creates value #669


odahcam:
> **Disclaimer:** this issue was originally submitted here: https://github.com/PHP-DI/PHP-DI/issues/669 . The context is about PHP-DI.

When the class gets destroyed in its life cycle, the `__destruct` method gets called and, if there's one implementation in the original class, the proxy class tries to execute the base class implementation. But looks like the proxy class is getting destroyed before creating a instance for the base class and, as there's no check for null values, we get the following error:

```
[Wed Jun 12 00:52:18 2019] ::1:59011 [500]: / - Uncaught Error: Call to a member function __destruct() on null in <Project-path>\var\proxies\ProxyManagerGeneratedProxy__PM__<CLASS_NAME>Generated8d9f8c89dc680a18cbf6f65f21952d52.php:55
Stack trace:
#0 <Project-path>\vendor\php-di\php-di\src\Proxy\ProxyFactory.php(60): ProxyManagerGeneratedProxy\__PM__\<CLASS_NAME>\Generated8d9f8c89dc680a18cbf6f65f21952d52->__destruct()
#1 <Project-path>\vendor\php-di\php-di\src\Proxy\ProxyFactory.php(74): DI\Proxy\ProxyFactory->createProxy('<CLASS_NAME>', Object(Closure))
#2 <Project-path>\vendor\php-di\php-di\src\Compiler\ObjectCreationCompiler.php(140): DI\Proxy\ProxyFactory->generateProxyClass('<CLASS_NAME>')
#3 <Project-path>\vendor\php-di\php-di\src\Compiler\ObjectCreationCompiler.php(39): DI\Compiler\Obje in <Project-path>\var\proxies\ProxyManagerGeneratedProxy__PM_<CLASS_NAME>Generated8d9f8c89dc680a18cbf6f65f21952d52.php on line 55
```

The project path and class full qualified names were hidden because it's a private project, where the class is related to database connectivity.

That class just access authorization data and creates a new connection instance. The `__destruct` method implementation exists to ensure the connection will be closed when the class gets destroyed. 

In this scenario the class was never used and that's why I believe it's getting destructed before creating an instance for the original class.

Ocramius:
This is very likely a duplicate of https://github.com/Ocramius/ProxyManager/issues/373

odahcam:
[...]

Issue #373, opened 2017-06-09
Proxy->__destruct called although service's __construct() was never called


cjost1988:
Hey, 

as I can see there were some issues regarding the `proxy->__destruct()` reported before. I am running into the same issues like mentioned there.

An **external** service establishes connections on instance construction. Everything is fine until the proxy->__destruct() is invoked that itself forces instantiation of the service although it's not needed.

Is there any specific reason that you force instantiation of the proxied service on __destruct()? 

In my point of view that means additional performance costs that are not wanted - just creating the instance reserves additional memory. With this behavior you force the developers to not load data or establish connections on construction if they want to avoid the addition performance overhead.

cheers,
Chris

Ocramius:
Linking https://github.com/Ocramius/ProxyManager/issues/258
Linking https://github.com/Ocramius/ProxyManager/issues/262

 > Is there any specific reason that you force instantiation of the proxied service on __destruct()?

As already described in the issues above, this library cannot make any assumptions about the behavior of `__destruct`. Specifically, `__destruct` may be running critical parts of your logic, and therefore cannot be "skipped" or "ignored" by this library.

 > In my point of view that means additional performance costs that are not wanted - just creating the instance reserves additional memory. With this behavior you force the developers to not load data or establish connections on construction if they want to avoid the addition performance overhead.

I don't force anything - the library is agnostic, and it will try honoring the initial behavior as much as possible.

As a workaround, I suggest implementing an intermediate class that doesn't implement `__destruct`, and all it does is forwarding method calls: you can then proxy that one class just to remove `__destruct` from the signature, and therefore from the initialization at engine shutdown.

Ocramius:
Just adding an example of how this would work, since this hackery would be EXTREMELY complex to add at ProxyManager level:

```php
class MyRedis
{
    public function __construct(string $dsn) { /* connect to redis here */ }
    public function doSomething() { /* interact with redis here */ }
    public function __destruct() { /* disconnect here */ }
}

class MyRedisWithoutSillyConstruct extends MyRedis
{
    public function __construct(MyRedis $originalInstance) { $this->originalInstance = $originalInstance; }
    public function doSomething() { return $this->originalInstance->doSomething(); }
    // no constructor
}

$lazyService = $factory->createProxy(MyRedisWithoutSillyConstruct::class, ...);
```

cjost1988:
Yes, I understand your point. But __destruct() makes only sense when the instance was constructed before?

[...]

PR #641, opened 2020-12-18
Add "skipDestructor" option to lazy loading proxies

nicolas-grekas:
This PR adds a new `skipDestructor` option to lazy proxies (ghosts and value holders) that skips calling the original destructor when the object is not initialized.

This is something that is much needed in DI containers, where e.g. we don't want opening a database connection on an unused Doctrine object.

For the record, Symfony already has a similar check in place, but it is implemented using monkey-patching. Because of the policies in place here (new features are available only on the very latest PHP version), I had absolutely no incentive in contributing here since I couldn't actually use that work anytime soon.

Now that I decided to [maintain a fork of this code](https://github.com/Ocramius/ProxyManager/issues/630#issuecomment-746158099), I know that anything I contribute here could be used in the short term.  That's a great incentive.

The good news is that you don't have to change your policies. OSS dynamics FTW here also.

### DIFF
--- a/src/ProxyManager/ProxyGenerator/LazyLoadingGhostGenerator.php
+++ b/src/ProxyManager/ProxyGenerator/LazyLoadingGhostGenerator.php
@@ -25,6 +25,7 @@ use ProxyManager\ProxyGenerator\LazyLoadingGhost\MethodGenerator\MagicSet;
 use ProxyManager\ProxyGenerator\LazyLoadingGhost\MethodGenerator\MagicSleep;
 use ProxyManager\ProxyGenerator\LazyLoadingGhost\MethodGenerator\MagicUnset;
 use ProxyManager\ProxyGenerator\LazyLoadingGhost\MethodGenerator\SetProxyInitializer;
+use ProxyManager\ProxyGenerator\LazyLoadingGhost\MethodGenerator\SkipDestructor;
 use ProxyManager\ProxyGenerator\LazyLoadingGhost\PropertyGenerator\InitializationTracker;
 use ProxyManager\ProxyGenerator\LazyLoadingGhost\PropertyGenerator\InitializerProperty;
 use ProxyManager\ProxyGenerator\LazyLoadingGhost\PropertyGenerator\PrivatePropertiesMap;
@@ -48,7 +49,7 @@ class LazyLoadingGhostGenerator implements ProxyGeneratorInterface
     /**
      * {@inheritDoc}
      *
-     * @psalm-param array{skippedProperties?: array<int, string>} $proxyOptions
+     * @psalm-param array{skippedProperties?: array<int, string>, skipDestructor?: bool} $proxyOptions
      *
      * @return void
      *
@@ -65,6 +66,7 @@ class LazyLoadingGhostGenerator implements ProxyGeneratorInterface
         $publicProperties    = new PublicPropertiesMap($filteredProperties);
         $privateProperties   = new PrivatePropertiesMap($filteredProperties);
         $protectedProperties = new ProtectedPropertiesMap($filteredProperties);
+        $skipDestructor      = ($proxyOptions['skipDestructor'] ?? false) && $originalClass->hasMethod('__destruct');
 
         $classGenerator->setExtendedClass($originalClass->getName());
         $classGenerator->setImplementedInterfaces([GhostObjectInterface::class]);
@@ -81,7 +83,7 @@ class LazyLoadingGhostGenerator implements ProxyGeneratorInterface
                 ClassGeneratorUtils::addMethodIfNotFinal($originalClass, $classGenerator, $generatedMethod);
             },
             array_merge(
-                $this->getAbstractProxiedMethods($originalClass),
+                $this->getAbstractProxiedMethods($originalClass, $skipDestructor),
                 [
                     $init,
                     new StaticProxyConstructor($initializer, $filteredProperties),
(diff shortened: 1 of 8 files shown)


src/ProxyManager/ProxyGenerator/LazyLoadingValueHolderGenerator.php
```diff
@@ -52,8 +53,10 @@ class LazyLoadingValueHolderGenerator implements ProxyGeneratorInterface
      *
      * @throws InvalidProxiedClassException
      * @throws InvalidArgumentException
+     *
+     * @psalm-param array{skipDestructor?: bool} $proxyOptions
      */
-    public function generate(ReflectionClass $originalClass, ClassGenerator $classGenerator)
+    public function generate(ReflectionClass $originalClass, ClassGenerator $classGenerator, array $proxyOptions = [])
```
nicolas-grekas, line 59: Because the class is not final, you may consider this as a BC break.
The solution I know about is to use `func_get_arg()` alongside an `@param` annotation that tells about the extra argument. If worth it, we could also trigger a runtime deprecation notice, that would be triggered when a child class doesn't declare the new argument.

Let me know how you'd like to deal with this.

nicolas-grekas:
About the mutation score, I tried the mutants locally, they all fail when applied. I don't know what is missing, please advise.

src/ProxyManager/ProxyGenerator/LazyLoadingValueHolderGenerator.php
alcaeus: Using func_get_args in combination with a psalm-var annotation would make sense. IIRC this package does not announce deprecations via trigger_error, but you could do that in the lts package if so inclined.

src/ProxyManager/ProxyGenerator/LazyLoadingGhostGenerator.php
```diff
@@ -134,8 +137,14 @@ static function (MethodGenerator $generatedMethod) use ($originalClass, $classGe
      *
      * @return MethodGenerator[]
      */
-    private function getAbstractProxiedMethods(ReflectionClass $originalClass): array
+    private function getAbstractProxiedMethods(ReflectionClass $originalClass, bool $skipDestructor): array
     {
+        $excludedMethods = ProxiedMethodsFilter::DEFAULT_EXCLUDED;
+
+        if ($skipDestructor) {
+            $excludedMethods[] = '__destruct';
```
alcaeus, line 153: Nit: rename the option to "excludeDestructor" since that seems to be the internal wording.

alcaeus (approved):
Nitpicking so @ocramius doesn't need to point it out, but in favour of the change 👍

Ocramius:
Potentially, this could also be a solution to #468 and #373, and giving the end user the foot-gun seems like a sensible choice.

Patch makes sense, although I do need to take more time to review.

Be aware that this will only land in a new minor release, which may end up being `~8.0.0`, or perhaps I do a new minor with just this, then one with `~8.0.0` if it helps.

src/ProxyManager/ProxyGenerator/LazyLoadingGhostGenerator.php
nicolas-grekas: Not sure about this proposal: "exclude" here is used for a different thing.

src/ProxyManager/ProxyGenerator/LazyLoadingValueHolderGenerator.php
nicolas-grekas: BC break now addressed using a few comments/annotations.

src/ProxyManager/ProxyGenerator/LazyLoadingGhostGenerator.php
alcaeus: My thinking was that you're excluding the destructor from the list of proxied methods, but I defer to your (and @Ocramius') judgment on this 👍

src/ProxyManager/ProxyGenerator/Util/ProxiedMethodsFilter.php
```diff
@@ -19,8 +19,7 @@
  */
 final class ProxiedMethodsFilter
 {
-    /** @var array<int, string> */
-    private static array $defaultExcluded = [
+    public const DEFAULT_EXCLUDED = [
```
Ocramius: Expanding the public API here is to be avoided

src/ProxyManager/ProxyGenerator/LazyLoadingValueHolderGenerator.php
```diff
@@ -52,9 +55,14 @@ class LazyLoadingValueHolderGenerator implements ProxyGeneratorInterface
...
      * @throws InvalidProxiedClassException
      * @throws InvalidArgumentException
+     *
+     * @psalm-param array{skipDestructor?: bool} $proxyOptions
      */
-    public function generate(ReflectionClass $originalClass, ClassGenerator $classGenerator)
+    public function generate(ReflectionClass $originalClass, ClassGenerator $classGenerator/*, array $proxyOptions = []*/)
     {
+        /** @psalm-var array{skipDestructor?: bool} $proxyOptions */
+        $proxyOptions = func_num_args() >= 3 ? func_get_arg(2) : [];
```
Ocramius: IMO BC break is a better approach here, and preferable to this sort of API design.

Ocramius: This one is also tripping the mutation test, as far as I can see

Ocramius:
No time to get it in `2.11.0`, hence moving it to `2.12.0`

nicolas-grekas:
Rebased, let me know the next iteration.

PR #642, opened 2020-12-19
Replace fatal error with notice when accessing proxied interface dynamic property

nicolas-grekas:
Submitting #419 again, with the last comments there fixed.

### DIFF
--- a/src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/MagicGet.php
+++ b/src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/MagicGet.php
@@ -38,7 +38,8 @@ class MagicGet extends MagicMethodGenerator
                 'name',
                 null,
                 null,
-                'returnValue'
+                'returnValue',
+                $originalClass->isInterface() ? $originalClass->getName() : null
             );
         }
 
--- a/src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
+++ b/src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
@@ -8,6 +8,9 @@ use InvalidArgumentException;
 use Laminas\Code\Generator\PropertyGenerator;
 
 use function sprintf;
+use function str_replace;
+use function substr;
+use function var_export;
 
 /**
  * Generates code necessary to simulate a fatal error in case of unauthorized
@@ -33,6 +36,7 @@ class PublicScopeSimulator
      *                                              to read the property. `$this` if none provided
      * @param string|null       $returnPropertyName name of the property to which we want to assign the result of
      *                                              the operation. Return directly if none provided
+     * @param string|null       $interfaceName      name of the proxified interface if any
      *
      * @throws InvalidArgumentException
      */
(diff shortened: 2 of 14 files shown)


Ocramius:
I don't think this should be supported, since interfaces have no `public` properties, and what you get from the factory is a `Proxy&YourInterface`, rather than a `Proxy&YourConcreteImplementation` (at type level).

For example, https://github.com/InterNations/type-jail uses `__get` and `__set` to explicitly deny access to anything not defined at interface level, and it's done through this restriction on this package, effectively :thinking:

Ocramius:
Hmm, unsure why https://github.com/Ocramius/ProxyManager/pull/419#event-2548105116 was closed - looks like an event triggered that? Totally happened below my radar, sorry :-\

src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
```diff
@@ -73,23 +84,29 @@ public static function getPublicAccessSimulationCode(
...
             return '';
         }
 
-        return '    $backtrace = debug_backtrace(false, 1);' . "\n"
+        $code = '    $backtrace = debug_backtrace(false, 1);' . "\n"
             . '    trigger_error(' . "\n"
             . '        sprintf(' . "\n"
             . '            \'Undefined property: %s::$%s in %s on line %s\',' . "\n"
-            . '            get_parent_class($this),' . "\n"
+            . '            ' . ($interfaceName !== null ? var_export($interfaceName, true) : 'get_parent_class($this)') . ',' . "\n"
```
Ocramius, line 97: Ah, wait, so this just produces a better notice for the interface?

src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
```diff
@@ -51,6 +56,12 @@ public static function getPublicAccessSimulationCode(
             $target = '$this->' . $valueHolder->getName();
         }
 
+        if ($interfaceName !== null) {
+            return '$targetObject = ' . $target . ';' . "\n\n"
+                . self::getUndefinedPropertyNotice($operationType, $nameParameter, $interfaceName)
+                . self::getOperation($operationType, $nameParameter, $valueParameter);
+        }
+
         return '$realInstanceReflection = new \\ReflectionClass(get_parent_class($this));' . "\n\n"
```
nicolas-grekas, line 65: without this change, this code is hit: `get_parent_class($this)` returns `false`, which is cast to the empty string by `ReflectionClass`, and a fatal error occurs: `Class  does not exist`.

src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php
nicolas-grekas: this removes a fatal error (see my other comment) and turns it into a proper notice (which also tells "don't do this!", but in a better way than a fatal error)

src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/MagicGet.php
```diff
@@ -38,7 +38,8 @@ public function __construct(
                 'name',
                 null,
                 null,
-                'returnValue'
+                'returnValue',
+                $originalClass->isInterface() ? $originalClass->getName() : null
```
Ocramius, line 42: Mutation test score went down: that's because each of these magic methods requires testing for generated code in case of interfaces given

Ocramius: I've picked up the branch and am refactoring now.

Ocramius:
Handled in #651, thanks @nicolas-grekas!

PR #643, opened 2020-12-19
Deprecate "Version" class, remove dep on "composer-runtime-api"

nicolas-grekas:
Same as #631 with a possible working implementation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,6 @@
     "require": {
         "php":                       "~7.4.1",
         "laminas/laminas-code":      "^3.4.1",
-        "composer-runtime-api":      "^2.0.0",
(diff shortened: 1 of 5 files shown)


Ocramius:
Closing as per #631 discussion: requiring `composer-runtime-api:^2` is simple/straightforward and well tested, and conditional dependencies are not acceptable from my end.

Besides, past versions of `ocramius/proxy-manager` work just fine with older versions of `ocramius/package-versions`, which didn't require `composer/composer:^2` implicitly.

This calls for:

 * adding https://github.com/maglnet/ComposerRequireChecker to the lot
 * getting rid of the outdated `laminas/laminas-rpc` components

I'll probably get to work on those points at a later point in time.

Issue #632, opened 2020-11-17
Notice: Only variable references should be returned by reference


greg0ire:
**Description**  
I'm getting the notice in the title. The code that causes it looks like this:

```php
$accessor = function & () use ($targetObject, $name, $value) {
    return $targetObject->$name = $value;
};
```

**How to reproduce**  

I don't know how to reproduce the generation of this piece of code yet, but reproducing the notice is easy enough:

https://3v4l.org/dE26T

I suspect it is related to this service definition (which looks quite simple):

```yaml
    #Config amqp_lib
    PhpAmqpLib\Connection\AMQPStreamConnection:
        lazy: true
        class: PhpAmqpLib\Connection\AMQPStreamConnection
        arguments:
            - "%swarrot.host%"
            - "%swarrot.port%"
            - "%swarrot.login%"
            - "%swarrot.password%"
            - "%swarrot.vhost%"
```

**Possible solution**

Do it in 2 steps: https://3v4l.org/j501X ?

Stack trace: 

```

/var/www/var/cache/prod/ContainerMifQx3V/AMQPStreamConnection_99651dd.php:251,
--
/var/www/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php:643,
/var/www/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Channel/AbstractChannel.php:234,
/var/www/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Channel/AbstractChannel.php:351,
/var/www/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Channel/AMQPChannel.php:287,
/var/www/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Channel/AMQPChannel.php:117,
/var/www/var/cache/prod/ContainerMifQx3V/getAMQPChannelService.php:35,
/var/www/var/cache/prod/ContainerMifQx3V/getAMQPChannelService.php:23,
/var/www/var/cache/prod/ContainerMifQx3V/AMQPChannel_f219a5d.php:116,
/var/www/var/cache/prod/ContainerMifQx3V/AMQPChannel_f219a5d.php:116,
/var/www/src/Index/Prepare/PrepareProductToIndex.php:173,
/var/www/src/Index/Prepare/PrepareProductToIndex.php:114,
/var/www/src/Index/Prepare/PrepareProductToIndex.php:70,
[...]

PR #645, opened 2020-12-22
Fix "Only variable references should be returned by reference"

nicolas-grekas:
As explained in #632, and easily confirmed by reading the generated code.
I suppose you'd also like a test case that actually runs this code path?
I'm not sure I'm going to be able to provide one.

/cc @greg0ire in case you'd like to provide one?

Feel free to close of course if that's not good enough.


Ocramius:
> I suppose you'd also like a test case that actually runs this code path?

Yes, that's the bit that we didn't have in the current test suite, and hence why #632 went under the radar

nicolas-grekas:
Now with tests and more fixes.

Ocramius:
Reminder to self: retarget to `2.10.x` before merging

nicolas-grekas:
> Reminder to self: retarget to 2.10.x before merging

done

nicolas-grekas:
Wrong push, my bad!

PR #647, opened 2020-12-22
Add "fluentSafe" option to LazyLoadingValueHolderGenerator

nicolas-grekas:
I propose to add a "fluentSafe" option to lazy value holders, to have them return the proxy object instead of the value holder when fluent methods are found.

This would be nice to use with interface-based lazy proxies where the proxy is used as a "condom" to prevent access to methods that are on the implementation but not on the proxied interface.

### DIFF
--- a/src/ProxyManager/Factory/AbstractBaseFactory.php
+++ b/src/ProxyManager/Factory/AbstractBaseFactory.php
@@ -17,6 +17,8 @@ use function array_key_exists;
 use function assert;
 use function class_exists;
 use function is_a;
+use function serialize;
+use function sha1;
 
 /**
  * Base factory common logic
@@ -29,7 +31,7 @@ abstract class AbstractBaseFactory
      * Cached checked class names
      *
      * @var array<string, string>
-     * @psalm-var array<class-string, class-string>
+     * @psalm-var array<string, class-string>
      */
     private array $checkedClasses = [];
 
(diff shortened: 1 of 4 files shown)


Ocramius (changes requested):
Don't think this is to be called "fluent safe".

There are two kinds of fluent interfaces:

 1. immutable API
 2. mutable API

In one case, we want to explicitly return the wrapped object (immutable), in the other, we generally want to return the proxy (mutable API).

In addition to that, nothing guarantees that an interface doesn't implement both flavors together.

Overall, my endorsement is to use a ["lazy loading ghost"](https://github.com/Ocramius/ProxyManager/blob/2.10.0/docs/lazy-loading-ghost-object.md) for anything that may implement fluent APIs, while a ["virtual proxy"/"lazy loading value holder"](https://github.com/Ocramius/ProxyManager/blob/2.10.0/docs/lazy-loading-value-holder.md) is simply not suitable for that purpose, and this is not solvable programmatically unless we inspect the AST of every parent method.

The best approach remains to return exactly what the return value of the wrapped value call is, since lazy-loading already happened at that point in time.


EDIT: to further elaborate

```php
class MutableCounter
{
    public int $counter = 0;
    public function add(): self
    {
        $this->counter += 1;

        return $this;
    }
}
```

```php
class ImmutableCounter
{
    public int $counter = 0;
    public function add(): self
    {
        $counter = clone $this;

        $counter->counter += 1;

        return $counter;
    }
}
```

Both of the above are valid fluent interfaces, and both are "safe". The effect of using a value holder proxy on both is that calling `MyProxyOfACounter#add()` will return the actual `MutableCounter` or `ImmutableCounter`, rather than the **same** instance, which is something that is correct/expected.

What could seem strange to the end-user is that `get_class((new MyProxyOfMutableCounter)->add()) !== MyProxyOfMutableCounter::class;`

Adding more options to adjust fluent API behavior is not really a solution to this design problem: I've documented that in more detail at https://ocramius.github.io/blog/fluent-interfaces-are-evil/ (https://github.com/Ocramius/ocramius.github.com/blob/a299d98f540bb3bfcf43e1306879d0615eea4ca0/source/_posts/2013-11-07-fluent-interfaces-are-evil.md).

nicolas-grekas:
> in the other, we generally want to return the proxy (mutable API).

This is the exact and only situation that this PR "fixes".
But thanks for the explanation, that's consistent.

Ocramius:
To be clear, most of the work put in ghost objects spawned from this initial identity problem with proxies 👍

PR #649, opened 2020-12-28
Fix lazy-loading-value-holder-internal-php-classes.phpt

nicolas-grekas:
On PHP 8, ext-phar declares argument 1 of the compress() methods as `int`.
Passing a string yields an error. This fixes it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     },
     "require-dev": {
         "ext-phar":                     "*",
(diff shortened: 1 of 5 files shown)


Ocramius (approved):
LGTM :+1:

Ocramius:
Manually merged to `2.10.x` via 8e85357b800125dea30b7083a6bf1800801ad98b

PR #660, opened 2021-01-10
Various small documentation cleanups

wouterj:
I've grouped each change by commit, I think the commits are quite self-explanatory.

As a first-time reader of this docs, I also think it's good to order the docs (e.g. by numbering the file names or having an `index.md` file). I'm however not the correct person to decide the order of the docs :) (I can only see that "download.md" is probably first).

---

Btw, I've left the links to classes with `master` intact. GitHub automatically redirects those to the default branch - which I think is perfect for these links.


Ocramius (approved):
LGTM, thanks @wouterj!

Issue #682, opened 2021-04-06
Cannot use non-nullable properties with AccessInterceptorScopeLocalizer


petr-buchin:
For now it's impossible to create proxies for objects that have non-nullable properties, even if that properties are being set in constructor.

Example:

```php
namespace Test;

use ProxyManager\Factory\AccessInterceptorScopeLocalizerFactory as ProxyFactory;

class Foo
{
    public function __construct(protected Bar $bar)
    {
    }
}

$factory = new ProxyFactory();
$foo = new Foo();

$factory->makeProxy($foo, [], []);
```

This code generates an exception:

```
[ProxyManager\Exception\UnsupportedProxiedClassException]                                                                                     
  Cannot create references for following properties of class Test\Foo: bar
```

Is this an expected behavior?

Ocramius:
Yes, that is expected, since proxymanager needs a default value to exist
for these properties.

petr-buchin:
@Ocramius as far as I see, this behavior can be fixed, if method `Properties::onlyNonReferenceableProperties()` will use constructor method reflection to see, what arguments does constructor have.

This way it can be sure that this properties will be initialized at the time of referencing them.

Do I understand that right?

If so, I could create a PR to address this issue.

Ocramius:
ProxyManager cannot arbitrarily assign a value to those properties, and
they cannot be unset, unless a default value can be assigned by the tool.

Hence you need to modify your ctor to be `__construct(private ?A $a =
null)`, or declare the property explicitly, like we did pre-php-8.0.

Note that PHP 8.0 constructor signatures are automatically promoted to
property declarations: inspecting the constructor has no effect.

petr-buchin:
> or declare the property explicitly, like we did pre-php-8.0.

@Ocramius I tried declaring property explicitly and having standard constructor with assigning this property a value, but that did not help

Ocramius:
Sorry, I wasn't clear enough: the property must be nullable **and** have a
default value of `null` declared:

```php
private ?A $a = null;
```

petr-buchin:
@Ocramius oh, that's too bad to declare all properties as nullable :(

But anyway, thank you very much for your help!

Ocramius:
Yeah, it's a limitation that has been in place since PHP 7.4.

Note that decorator proxies still operate well with that.

petr-buchyn:
[...]

PR #685, opened 2021-04-12
Support for non-nullable typed properties

### LINKED ISSUE #683: Allowing non-nullable properties for AccessInterceptorScopeLocalizer
This is a follow up on https://github.com/Ocramius/ProxyManager/issues/682.

After doing some research, I realized that enabling AccessInterceptorScopeLocalizer to work with non-nullable properties is pretty straightforward.

For now checking for non-accessible properties looks like follows in class `Properties`:

```php
return new self(array_filter($this->properties, static function (ReflectionProperty $property): bool {
    if (! $property->hasType()) {
        return false;
    }

    return ! array_key_exists(
        $property->getName(),
        // https://bugs.php.net/bug.php?id=77673
        array_flip(array_keys($property->getDeclaringClass()->getDefaultProperties()))
    );
}));
```

To fix this behavior and allow non-nullable properties on objects being proxied, this code can be changed to something like follows:

```php
return new self(array_filter($this->properties, static function (ReflectionProperty $property) use($realObject): bool {
    if (! $property->hasType()) {
        return false;
    }
    
    if ($property->getType()->allowsNull() && $property->getType()->hasDefaultValue()) {
        return false;
    }

    return ! $property->isInitialized($realObject);
}));
```

This code allows to proxy objects with non-nullable properties and moves responsibility of initializing properties to the user of this library.

However, the problem is that classes `Properties` and `BindProxyProperties` does not have access to the object being proxied, thus cannot use method `ReflectionProperty::isInitialized()`.

It's possible to pass the instance of an object being proxied to this classes, but that would require changing signatures of some interface methods, and that would be a BC break.

I was thinking about using a temporary workaround (before next major release), such as `ProxyContext` class with static property, like `ProxyContext::$objectBeingProxied`, that would allow classes like `BindProxyProperties` to have access to the object being proxied without changing interfaces, but that is admittedly clunky. 

@Ocramius do you have any thoughts on this?

petr-buchin:
Fixes https://github.com/Ocramius/ProxyManager/issues/683 and https://github.com/Ocramius/ProxyManager/issues/682

This PR adds support for non-nullable typed properties when using AccessInterceptorScopeLocalizer proxy.

@Ocramius I tried to start from tests, as you said in https://github.com/Ocramius/ProxyManager/issues/683#issuecomment-815052652

> do start from the test suite, by adding a class that looks like what you are trying to proxy, and making it work from there

I added class `ClassWithNonNullableTypedProperties`, but as far as I see, existing tests are pretty informative already when used with this class, since all I want to achieve is to have this class proxied (which is being done in tests) and to have proxy being able to act like an original object (being in sync with it), which is also covered by existing tests.

So if you can think of any other tests that I need to write for this PR, please request changes and I'll do it.

Thanks in advance!


tests/ProxyManagerTest/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyPropertiesTest.php
```diff
@@ -69,30 +69,16 @@ public function testBodyStructure(): void
...
-\Closure::bind(function () use ($localizedObject) {
-    $this->privateProperty1 = & $localizedObject->privateProperty1;
-}, $this, 'ProxyManagerTestAsset\\ClassWithMixedProperties')->__invoke();
-
-\Closure::bind(function () use ($localizedObject) {
-    $this->privateProperty2 = & $localizedObject->privateProperty2;
-}, $this, 'ProxyManagerTestAsset\\ClassWithMixedProperties')->__invoke();
-
+$class = new \ReflectionObject($localizedObject);
+$this->bindProxyProperty($localizedObject, $class, 'publicProperty0');
```
Ocramius: It would be better to inline these, instead of generating `$this->bindProxyProperty()`.

Remember that this is generated code, so we don't really want to abstract it, but rather let the optimizer do as much work as possible.

In additon to that, `bindProxyProperty` is a too generic name, and will likely collide with parent class implementation.

src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyProperty.php
```diff
@@ -0,0 +1,65 @@
...
+                new ParameterGenerator('localizedObject', $originalClass->getName()),
+                new ParameterGenerator('class', ReflectionClass::class),
+                new ParameterGenerator('propertyName', 'string'),
+            ],
+            self::FLAG_PRIVATE
+        );
+
+        $bodyTemplate = <<<'CODE'
+        $originalClass = $class;
+        while (! $class->hasProperty($propertyName)) {
```
Ocramius: This operation should be performed during compilation, rather than during runtime: it's quite expensive to traverse ancestors.

src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyProperty.php
```diff
@@ -0,0 +1,65 @@
...
+            $class = $class->getParentClass();
+        }
+        $property = $class->getProperty($propertyName);
+        $property->setAccessible(true);
+        if (!$property->isInitialized($localizedObject)) {
```
Ocramius: If I understand this correctly, the difference here is that we aren't creating an "empty shell", but proxying when the other class is correctly initialized?

src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyProperty.php
```diff
@@ -0,0 +1,65 @@
...
+            throw new \{ EXCEPTION_CLASS }(
+                sprintf(
+                    'Cannot create reference for property $%s of class %s: property must be initialized',
+                    $property->getName(),
+                    $originalClass->getName()
+                )
+            );
+        }
+        if (!$property->isPrivate()) {
```
Ocramius: This call can be prevented by inspecting the reflection information for a property upfront

tests/ProxyManagerTest/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyPropertiesTest.php
```diff
@@ -198,116 +154,53 @@ public function testBodyStructureWithTypedProperties(): void
...
-\Closure::bind(function () use ($localizedObject) {
-    $this->privateIterableProperty = & $localizedObject->privateIterableProperty;
-}, $this, 'ProxyManagerTestAsset\\ClassWithMixedReferenceableTypedProperties')->__invoke();
-
-\Closure::bind(function () use ($localizedObject) {
-    $this->privateNullableIterableProperty = & $localizedObject->privateNullableIterableProperty;
-}, $this, 'ProxyManagerTestAsset\\ClassWithMixedReferenceableTypedProperties')->__invoke();
-
+        $expectedCode = <<<'PHP'
+$class = new \ReflectionObject($localizedObject);
```
Ocramius: Ideally, we should be able to avoid runtime reflection for all properties that do not have a declared type or do have a default `null` value

Ocramius (changes requested):

src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyProperty.php
petr-buchin: @Ocramius unfortunately I don't understand how to perform this during compilation: different properties would have different declaring classes. So how this can be done during compilation?

src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyProperty.php
petr-buchin: Yes: for now proxy can be created only when properties are nullable and have default value in class itself.
Thus this library tries to be responsible for user classes and objects.
This PR moves responsibility of initializing object to the user of the library: class may have non-nullable typed properties, but as long as concrete instance being proxied is properly initialized - it can be proxied.

petr-buchin:
@Ocramius seems like in the last commit I've adressed all issues you mentioned.

Could you please give it a new review?

Thanks!

tests/ProxyManagerTest/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyPropertiesTest.php
petr-buchin: Done

src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyProperty.php
petr-buchin: Done - I'm generating a `$classesMap` variable at build time, which has instances of `ReflectionClass` for private properties, and then use it when localizing private props.

src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyProperty.php
petr-buchin: Done.

tests/ProxyManagerTest/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyPropertiesTest.php
petr-buchin: Done - I've returned original version of code generation for untyped properties and nullable properties with default values.

Other properties (`$nonReferenceableProperties`) are then used to generate localization code with `ReflectionProperty::isInitialized()` checks.

petr-buchin:
@Ocramius sorry for bothering you, but will you have time to review this PR in the coming weeks?

Ocramius:
@petr-buchin possibly throwing it at the weekend - this week is packed with work.

petr-buchin:
@Ocramius thank you very much - it would be great to have this reviewed in the coming weeks whenever you have time for it.
So no hurry.

petr-buchin:
@Ocramius may I ask you to look into this when you have time?

We have entire framework depending on this feature, and it would be great to use this library instead of fork (as we do now).

Thanks in advance!

Ocramius:
Indeed, I should try and verify this locally, but did also not get to it as I was hoping, sorry :(

petr-buchin:
Dear @Ocramius, is there any chance you can look at this PR ?

Our company uses [fork](https://github.com/petr-buchin/ProxyManager/tree/non-nullable-typed-properties-support) of ProxyManager repo, that includes this PR, for almost a year.

Our 2 opensource repositories are dependent on this fork functionality.
Our clients are the biggest banks in a world, so this code is "battle-tested" for almost a year in production and did not cause any problems or performance-related troubles.

But now we need to install `symfony/orm-pack`, which depends on `friendsofphp/proxy-manager-lts`, which is a fork from your repository.

`friendsofphp/proxy-manager-lts` replaces `ocramius/ProxyManager` lib via it's Composer settings, and your repository is upstream for it, thus to use it, I must have this PR merged here and updated from upstream in `friendsofphp/proxy-manager-lts`.

Could you please look at this PR? Maybe I can do something for you to consider this PR?

Thanks in advance!

Ocramius:
Hey @petr-buchin, I'll put it on the milestone for the next minor, which includes PHP 8.1.x support, but I haven't actively worked on `ocramius/proxy-manager`, other than bugfixes, for a while now. 

As for the clients using this, I don't really care much, an mentioning that they're the biggest banks in the world only reduces my will to throw unpaid time at it :D 

I'll see when I can work on this :)

petr-buchin:
@Ocramius I do understand you :)

If it matters - I don't work for a bank, banks are clients of company where I work :)

> an mentioning that they're the biggest banks in the world only reduces my will to throw unpaid time at it

Сan I pay for you time to look at this?

If I do, please email me at `petrbuchyn@gmail.com`.

Thanks very much!

composer.json
```diff
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "php":                       "~8.0.0",
+        "php":                       ">=8.0.0",
```
Ocramius, line 22: Should be `~8.0.0`

petr-buchin: Fixed

nicolas-grekas:
It'd be great rebasing + squashing this PR in one commit on top of 2.15.x

src/ProxyManager/ProxyGenerator/AccessInterceptorScopeLocalizer/MethodGenerator/BindProxyProperties.php
```diff
@@ -70,10 +65,101 @@ public function __construct(
...
+            ? sprintf(
+                '$property = $classesMap[\'%s\']->getProperty(\'%s\');',
+                $declaringClassName,
+                $property->getName()
+            )
+            : sprintf('$property = $class->getProperty(\'%s\');', $property->getName());
+        $localizedPropertyCode .= PHP_EOL;
+
+        if (! $property->isPublic()) {
+            $localizedPropertyCode .= '$property->setAccessible(true);' . PHP_EOL;
```
nicolas-grekas, line 127: can be removed since https://wiki.php.net/rfc/make-reflection-setaccessible-no-op

nicolas-grekas (commented):
Shouldn't this be able to support readonly properties also?
Since they are readonly, using a reference to bind instances is not possible, but copying the value of these properties in the proxy would be just fine, since... they would be readonly :)

ErvinSabic:
We're running into a support requirement for this as well and will be using this PR branch. But it would be awesome if it could be merged in directly :smile:

PR #689, opened 2021-04-19
Remove redundant error handler in generate

michalbundyra:
Relates to webimpress/safe-writer#49

The problem is with testing in ocramius/proxy-manager, not in safe-writer itself.

Update: still it can be improved in safe-writer to set custom error-handler instead of suppressing warnings with `@`.

### DIFF
--- a/src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
+++ b/src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
@@ -4,15 +4,11 @@ declare(strict_types=1);
 
 namespace ProxyManager\GeneratorStrategy;
 
-use Closure;
 use Laminas\Code\Generator\ClassGenerator;
 use ProxyManager\Exception\FileNotWritableException;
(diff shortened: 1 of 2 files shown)


Ocramius:
Aha, I guess the `@` part really was the catch :+1: 

Still, won't that trip tools like sentry, which declare custom error handlers to catch these cases? :thinking:

src/ProxyManager/GeneratorStrategy/FileWriterGeneratorStrategy.php
```diff
@@ -44,16 +37,12 @@ public function generate(ClassGenerator $classGenerator): string
         $className     = $classGenerator->getNamespaceName() . '\\' . $classGenerator->getName();
         $fileName      = $this->fileLocator->getProxyFileName($className);
 
-        set_error_handler($this->emptyErrorHandler);
```
Ocramius, line 47: Discussed in chat: change is valid, I just want to merge the removal of the error handlers separately from the test adjustment first, but overall 👍

Ocramius (approved):

PR #700, opened 2021-05-21
Make library compatible with `composer/composer:^2.0.14`, do not call `InstalledVersions::getRawData()` (which introduces runtime side-effect in pure API)

derrabus:
Follows composer/composer#9816

This PR proposes to not call `InstalledVersions::getRawData()` anymore. This method has been deprecated and triggers runtime deprecation warnings since Composer 2.0.14.

### DIFF
--- a/src/ProxyManager/Generator/Util/IdentifierSuffixer.php
+++ b/src/ProxyManager/Generator/Util/IdentifierSuffixer.php
@@ -6,7 +6,6 @@ namespace ProxyManager\Generator\Util;
 
 use Composer\InstalledVersions;
 
-use function is_callable;
 use function preg_match;
 use function serialize;
 use function sha1;
@@ -49,10 +48,6 @@ abstract class IdentifierSuffixer
 
     private static function loadBaseHashSalt(): string
     {
-        return sha1(serialize(
-            is_callable([InstalledVersions::class, 'getAllRawData'])
-                ? InstalledVersions::getAllRawData() // Composer >= 2.0.14
-                : InstalledVersions::getRawData()
-        ));
+        return sha1(serialize(InstalledVersions::getAllRawData()));
     }
 }
--- a/src/ProxyManager/Generator/Util/IdentifierSuffixer.php
+++ b/src/ProxyManager/Generator/Util/IdentifierSuffixer.php
@@ -6,6 +6,7 @@ namespace ProxyManager\Generator\Util;
 
 use Composer\InstalledVersions;
 
+use function is_callable;
 use function preg_match;
 use function serialize;
 use function sha1;
@@ -48,6 +49,10 @@ abstract class IdentifierSuffixer
 
     private static function loadBaseHashSalt(): string
     {
-        return sha1(serialize(InstalledVersions::getRawData()));
+        return sha1(serialize(
+            is_callable([InstalledVersions::class, 'getAllRawData'])
+                ? InstalledVersions::getAllRawData() // Composer >= 2.0.14
+                : InstalledVersions::getRawData()
+        ));
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "webimpress/safe-writer":    "^2.2.0"
     },
     "conflict": {
+        "composer/composer":         "<2.0.14",
         "zendframework/zend-stdlib": "<3.2.1",
         "laminas/laminas-stdlib":    "<3.2.1",
         "doctrine/annotations":      "<1.6.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "1c17c70d4578784a835b559dbec85ee2",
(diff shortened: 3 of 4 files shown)


Ocramius:
Is there a way to put a constraint on `"composer-runtime-api": "^2.0.14"`
instead?

derrabus:
I tried to do that, but apparently Composer 2.0.14 still claims to provide `"composer-runtime-api": "2.0.0"`.

derrabus:
```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires composer-runtime-api ^2.0.14, found composer-runtime-api[2.0.0] but it does not match the constraint.
```

Ocramius:
I wonder if we should conflict with `composer/composer:<2.0.14` then 🤔

Overall, feels weird that the plugin API doesn't reflect composer version:
we should probably report that in upstream.

Also, runtime deprecation warnings are reeeeeeally a no-go 😩

derrabus:
cc @Seldaek

Seldaek:
I was pretty sure nobody relied on this method but of course @Ocramius to the rescue :p

Anyway yes should have bumped the runtime api on this one sorry. 2.1 may come out soon if i manage tho, so we'll see what makes it first..

That said IMO merging this as is would be better as it ensures broader compatibility and people are notoriously bad at keeping composer up to date..

src/ProxyManager/Generator/Util/IdentifierSuffixer.php
```diff
@@ -48,6 +49,10 @@ public static function getIdentifier(string $name): string
 
     private static function loadBaseHashSalt(): string
     {
-        return sha1(serialize(InstalledVersions::getRawData()));
+        return sha1(serialize(
+            is_callable([InstalledVersions::class, 'getAllRawData'])
+                ? InstalledVersions::getAllRawData() // Composer >= 2.0.14
```
Ocramius: Note: I'll add a `"conflict": {"composer/composer": "<2.0.14"}` before merging, and remove the conditional.

Ocramius (approved):

src/ProxyManager/Generator/Util/IdentifierSuffixer.php
Ocramius: b5ee473

Ocramius:
Thanks @derrabus! Rolling a release nao :-)

Tobion:
This breaks every app using composer <2.0.14 or am I missing something?!

> bin/console c:c
Symfony\Component\ErrorHandler\Error\UndefinedMethodError^ {#13173
  #message: """
    Attempted to call an undefined method named "getAllRawData" of class "Composer\InstalledVersions".\n
    Did you mean to call "getRawData"?
    """
  #code: 0
  #file: "./vendor/ocramius/proxy-manager/src/ProxyManager/Generator/Util/IdentifierSuffixer.php"
  #line: 51
  trace: {
    ./vendor/ocramius/proxy-manager/src/ProxyManager/Generator/Util/IdentifierSuffixer.php:

Ocramius:
If that's the case, then `composer/composer` is not declaring itself at runtime, which is potentially yet another upstream bug/problem :shrug:

Seldaek:
composer/composer is not available as a platform package no.. composer-runtime-api and composer-plugin-api are. Otherwise it wouldn't really be possible to add composer/composer as a dependency and have it installed in the vendor dir. Not something we're going to fix.

Ocramius:
Noted: for now, I'll point people observing crashes to this discussion then.

I've raised #702 for when the new plugin API is available.

src/ProxyManager/Generator/Util/IdentifierSuffixer.php
```diff
@@ -48,6 +48,6 @@ public static function getIdentifier(string $name): string
 
     private static function loadBaseHashSalt(): string
     {
-        return sha1(serialize(InstalledVersions::getRawData()));
+        return sha1(serialize(InstalledVersions::getAllRawData()));
```
Tobion, line 51: Can't you add the `is_callable([InstalledVersions::class, 'getAllRawData'])` check here as well? This way, people won't run into this problem like I did.

Ocramius: https://github.com/Ocramius/ProxyManager/pull/700#discussion_r638675531

Tobion: yeah but it does not work

Venorcis:
> This breaks every app using composer <2.0.14 or am I missing something?!
> 
> > bin/console c:c
> > Symfony\Component\ErrorHandler\Error\UndefinedMethodError^ {#13173
> > #message: """
> > Attempted to call an undefined method named "getAllRawData" of class "Composer\InstalledVersions".\n
> > Did you mean to call "getRawData"?
> > """
> > #code: 0
> > #file: "./vendor/ocramius/proxy-manager/src/ProxyManager/Generator/Util/IdentifierSuffixer.php"
> > #line: 51
> > trace: {
> > ./vendor/ocramius/proxy-manager/src/ProxyManager/Generator/Util/IdentifierSuffixer.php:

I have composer 2.0.8 in my Docker instance and indeed this error pops up with 2.12.0
I'd suggest reverting the latest commit in master (keep the check on getRawData/getAllRawData)

PR #675, opened 2021-03-15
Bump vimeo/psalm from 4.4.1 to 4.6.3

malukenho:
Favor this PR over https://github.com/Ocramius/ProxyManager/pull/674

PR #711, opened 2021-07-13
Dependency upgrades and general cleanup

### LINKED ISSUE #167: Drop PHP 5.4 support

### LINKED ISSUE #338: Double call the parent method, when the name of the method matches the class name
I was playing around with proxies and suddenly encountered a strange issue, reproduced by the code below:

```
$proxy = new \ProxyManager\Factory\AccessInterceptorScopeLocalizerFactory();

class Foo
{
    public function foo()
    {
        echo '2';
    }
}
$foo = $proxy->createProxy(
    new Foo(),
    ['foo' => function() {echo '1';}],
    ['foo' => function () { echo '3';}]
);

$foo->foo();
```

Expected output: `123`
Actual output: `22`
Tested on: PHP 7.0.11
Affected Package Versions: `^1,0`, `^2.0` 

This is definitely not an expected behaviour, since it is not documented anywhere. I personally don't see it intended behaviour, am I wrong?

### LINKED ISSUE #659: Remove PHP 7.x support

### LINKED ISSUE #706: PHPUnit XSD should point at local reference under `vendor/`, and not at `https://schema.phpunit.de/`

Ocramius:
This patch:

 * drops PHP 7.x support
 * upgrades composer dependencies
 * upgrades coding standard

Closes #674 
Closes #167 
Closes #659 
Closes #706 
Closes #338 
Closes #675


src/ProxyManager/Factory/RemoteObject/AdapterInterface.php
```diff
@@ -13,8 +13,6 @@ interface AdapterInterface
      * Call remote object
      *
      * @param array<int, mixed> $params
-     *
-     * @return mixed
      */
-    public function call(string $wrappedClass, string $method, array $params = []);
+    public function call(string $wrappedClass, string $method, array $params = []): mixed;
```
Ocramius: This change is a BC break, because a child class declaring no return type is not considered compatible according to current PHP 8.x variance semantics

PR #717, opened 2021-09-01
Don't call deprecated InstalledVersions::getRawData()

KhadaRoss:
Logs get spammed with an error to not call getRawData() since its depricated.
I used the same fix as the 2.12.x branch instead of updating to it, since it requires PHP 8 and out application is still on 7.4.

### DIFF
--- a/src/ProxyManager/Generator/Util/IdentifierSuffixer.php
+++ b/src/ProxyManager/Generator/Util/IdentifierSuffixer.php
@@ -6,6 +6,7 @@ namespace ProxyManager\Generator\Util;
 
 use Composer\InstalledVersions;
 
+use function is_callable;
 use function preg_match;
 use function serialize;
 use function sha1;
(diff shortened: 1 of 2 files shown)


Ocramius:
Unless it's related to a security issue, past releases are not being
patched.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

KhadaRoss:
It is a huge barrier to update to 2.12.x since it's required to migrate the whole business application to PHP 8.
This is a low hanging fruit to fix an insane amount of spam in the logs for people still on PHP 7.4, without having to fork the ProxyManager on their own.
Maybe you can make an exception in this getRawData case.

Ocramius:
Generally not open for that: suppress deprecation warnings, or alternatively pin to a much older release that still uses `ocramius/package-versions`  (such as `ocramius/proxy-manager:2.8.*`). Adding more bugfixes would just complicate security issue bugfixing, should it happen.

See also https://github.com/Ocramius/.github/blob/6d4561515513f6d59fb144d838971d4a7715e3e4/version-support.md#support-policy

I would additionally suggest to pressure libraries that abuse `trigger_error()` for deprecations: that's an unhealthy practice that is developing further and further in PHP, and it's making overall maintenance work a living hell.

KhadaRoss:
I just realized that newer versions of the ProxyManager have ~7.4.1 || ~8.0.0 and not ~8.0.0 only.
My bad i will update the depencency. Thanks for your time.

PR #720, opened 2021-12-03
Update `https` instead of `http` in `composer.json` URLs

S-Tian86:
Fix Issue FriendsOfPHP/proxy-manager-lts#15


Ocramius (approved):
Thanks @S-Tian86!

PR #731, opened 2021-12-07
Added PHP 8.1 to CI pipeline

### LINKED ISSUE #721: Add PHP 8.1 to CI pipeline
See https://github.com/Ocramius/ProxyManager/blob/95f9ccf6711b91736ce6da261c24252b84a290c7/.github/workflows/phpunit.yml#L19-L20 (and similar files in the same directory)

We need to run everything also against PHP 8.1

michaelpetri:
Resolves https://github.com/Ocramius/ProxyManager/issues/721

Does it make sense to switch to https://github.com/laminas/laminas-continuous-integration-action ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "php":                       "~8.0.0",
+        "php":                       "~8.0.0 || ~8.0.0",
         "composer-runtime-api":      "^2.1.0",
         "laminas/laminas-code":      "^4.4.2",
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "php":                       "~8.0.0",
+        "php":                       "~8.0.0 || ~8.0.0",
```
Steveb-p:
```suggestion
        "php":                       "~8.0.0 || ~8.1.0",
```

PR #735, opened 2022-02-18
Fix code generation for `never` return type

### LINKED ISSUE #729: Verify code generation for the new `never` type
The test suite includes scenarios that handle the `: void` return type of methods: 

 * https://github.com/Ocramius/ProxyManager/blob/95f9ccf6711b91736ce6da261c24252b84a290c7/tests/ProxyManagerTestAsset/VoidCounter.php#L7-L17
 * https://github.com/Ocramius/ProxyManager/blob/95f9ccf6711b91736ce6da261c24252b84a290c7/tests/ProxyManagerTestAsset/VoidMethodTypeHintedClass.php
 * https://github.com/Ocramius/ProxyManager/blob/95f9ccf6711b91736ce6da261c24252b84a290c7/tests/ProxyManagerTestAsset/VoidMethodTypeHintedInterface.php

We need:

 1. similar classes, but using the `never` type ( https://wiki.php.net/rfc/noreturn_type )
 2. make sure we find usages of above classes, and introduce tests around the `never` type wherever they are used (using the newly introduced assets)

alekitto:
Fixes #729 

Based upon #731 to execute pipeline on PHP 8.1.
I've fixed the deprecations raised by implementing `\Serializable` without the new `__serialize/__unserialize` methods, but I needed to disable deprecation report on `lazy-loading-value-holder-internal-php-classes.phpt` due to `#[\ReturnTypeWillChange]`-related errors.

### DIFF
--- a/src/ProxyManager/Generator/Util/ProxiedMethodReturnExpression.php
+++ b/src/ProxyManager/Generator/Util/ProxiedMethodReturnExpression.php
@@ -16,14 +16,27 @@ final class ProxiedMethodReturnExpression
 {
     public static function generate(string $returnedValueExpression, ?ReflectionMethod $originalMethod): string
     {
-        $originalReturnType = $originalMethod === null
-            ? null
-            : $originalMethod->getReturnType();
+        $originalReturnType = $originalMethod?->getReturnType();
+        $expression         = null;
 
-        if ($originalReturnType instanceof ReflectionNamedType && $originalReturnType->getName() === 'void') {
+        if ($originalReturnType instanceof ReflectionNamedType) {
+            $expression = self::generateExpressionForNamedType($originalReturnType, $returnedValueExpression);
+        }
+
+        return $expression ?? 'return ' . $returnedValueExpression . ';';
+    }
+
+    private static function generateExpressionForNamedType(ReflectionNamedType $originalReturnType, string $returnedValueExpression): ?string
+    {
+        $originalReturnTypeName = $originalReturnType->getName();
+        if ($originalReturnTypeName === 'void') {
             return $returnedValueExpression . ";\nreturn;";
         }
 
-        return 'return ' . $returnedValueExpression . ';';
+        if ($originalReturnTypeName === 'never') {
+            return $returnedValueExpression . ';';
+        }
+
+        return null;
     }
 }
--- a/src/ProxyManager/ProxyGenerator/NullObject/MethodGenerator/NullObjectMethodInterceptor.php
+++ b/src/ProxyManager/ProxyGenerator/NullObject/MethodGenerator/NullObjectMethodInterceptor.php
@@ -7,7 +7,6 @@ namespace ProxyManager\ProxyGenerator\NullObject\MethodGenerator;
 use Laminas\Code\Reflection\MethodReflection;
 use ProxyManager\Generator\MethodGenerator;
 use ProxyManager\Generator\Util\IdentifierSuffixer;
-use ReflectionNamedType;
 
 /**
  * Method decorator for null objects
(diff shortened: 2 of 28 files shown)


tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php
```diff
@@ -1516,6 +1520,47 @@ static function (
...
+    /**
+     * @group 723
+     * @psalm-suppress UnusedVariable this method uses by-ref assignment of properties, and isn't recognized by static analysis
+     * @psalm-suppress UndefinedClass Class, interface or enum named never does not exist
+     */
+    public function testWillExecuteLogicInANeverMethod(): void
+    {
+        if (PHP_VERSION_ID < 80100) {
+            self::markTestSkipped('Needs PHP 8.1');
+        }
```
michaelpetri, line 1532: Can be replace by annotation:
```php
/** @requires PHP >= 8.1 */
```

alekitto: Similar checks have been used in other tests, but yes could be replaced easily if needed.

src/ProxyManager/ProxyGenerator/NullObject/MethodGenerator/NullObjectMethodInterceptor.php
```diff
@@ -18,9 +19,12 @@ class NullObjectMethodInterceptor extends MethodGenerator
      */
     public static function generateMethod(MethodReflection $originalMethod): self
     {
-        $method = static::fromReflectionWithoutBodyAndDocBlock($originalMethod);
+        $method             = static::fromReflectionWithoutBodyAndDocBlock($originalMethod);
+        $originalReturnType = $originalMethod->getReturnType();
 
-        if ($originalMethod->returnsReference()) {
+        if ($originalReturnType instanceof ReflectionNamedType && $originalReturnType->getName() === 'never') {
+            $method->setBody('throw new \Exception();');
```
nicolas-grekas: Is this really needed? without anything in the body, php already throws:
`TypeError: [...]: never-returning function must not implicitly return`

src/ProxyManager/Generator/Util/ProxiedMethodReturnExpression.php
```diff
@@ -16,14 +16,27 @@ final class ProxiedMethodReturnExpression
...
     public static function generate(string $returnedValueExpression, ?ReflectionMethod $originalMethod): string
     {
-        $originalReturnType = $originalMethod === null
-            ? null
-            : $originalMethod->getReturnType();
+        $originalReturnType = $originalMethod?->getReturnType();
+        $expression         = null;
 
-        if ($originalReturnType instanceof ReflectionNamedType && $originalReturnType->getName() === 'void') {
+        if ($originalReturnType instanceof ReflectionNamedType) {
```
nicolas-grekas, line 22: Wouldn't this diff be enough on this file?
```diff
-        if ($originalReturnType instanceof ReflectionNamedType && $originalReturnType->getName() === 'void') {
-            return $returnedValueExpression . ";\nreturn;";
+        if ($originalReturnType instanceof ReflectionNamedType && in_array($originalReturnType->getName(), ['void', 'never'])) {
+            return $returnedValueExpression . ';';
```

nicolas-grekas: Never mind, this doesn't work for `void` as there can be early returns.

src/ProxyManager/ProxyGenerator/NullObject/MethodGenerator/NullObjectMethodInterceptor.php
alekitto: Mh, probably is not needed. Don't rembember why I added it honestly, I'll remove that

PR #736, opened 2022-02-20
Support by-reference params in interceptors

ostrolucky:
This is needed in SncRedisBundle, see https://github.com/snc/SncRedisBundle/issues/651. We use interceptors there and currently all references are unset in proxies.


Ocramius:
Unit test is fine, but we should have an integration test here 🤔

ostrolucky:
I'm not familiar with code base, can you point me to an example testcase that you want to extend?

nicolas-grekas:
See phpt files in tests/language-feature-scripts/ @ostrolucky

Ocramius:
Also https://github.com/Ocramius/ProxyManager/blob/6d2458ddaf89f321ffb4a2539f17b83d761a0534/tests/ProxyManagerTest/Functional/AccessInterceptorValueHolderFunctionalTest.php

What's most useful here is having a test asset that reproduces the root issue somehow.

ostrolucky:
added

Ocramius (approved):
Thanks @ostrolucky! I converted the test into a regular PHPUnit test scenario, and fixed by-ref interceptors for both access-interceptor types meanwhile :)

Will cut a release now.

PR #738, opened 2022-03-01
Get ready for PHP 8.1

nicolas-grekas:
These test cases are needed to keep covering the related situations, but the deprecations thrown since PHP 8.1 should be silenced.

https://wiki.php.net/rfc/phase_out_serializable

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -15,7 +15,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "8.0"
+          - "8.1"
         operating-system:
           - "ubuntu-latest"
 
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -15,7 +15,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "8.0"
+          - "8.1"
         operating-system:
           - "ubuntu-latest"
 
--- a/.github/workflows/demo-scripts.yml
+++ b/.github/workflows/demo-scripts.yml
@@ -15,7 +15,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "8.0"
+          - "8.1"
         operating-system:
           - "ubuntu-latest"
 
--- a/.github/workflows/mutation-tests.yml
+++ b/.github/workflows/mutation-tests.yml
@@ -15,7 +15,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "8.0"
+          - "8.1"
         operating-system:
           - "ubuntu-latest"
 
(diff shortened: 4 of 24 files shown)


Ocramius (changes requested):
These should be in `--EXPECT--`.

The point of the `.phpt` suite is to expose every tiny issue with feature parity between native objects and ProxyManager doubles, which means that potential deprecations are part of it too.

Silencing deprecations is not viable, as it hides other (potentially worse) deprecations, which (as all deprecations) should be inspected and investigated as failures.

We'll also need to replicate all these tests with `Serializable` and `__serialize` once PHP 8.1 is added to CI.

nicolas-grekas:
I updated with PR with 3 commits:
- Update composer files for PHP 8.1
- Expect deprecations about Serializable
- Generate tentative types when appropriate

Together, the make tests green on PHP 8.1.

makaronnik:
Please speed up the update to 8.1, it's very necessary! 🙏

Ocramius:
I haven't worked on this yet: not on my priority list at the moment, as much more other stuff is in the way.

etremblay:
Would it be possible to at least lift the requirement for php ~8.0.0 in composer.json.  As I understand there is no actual risk for using the library with php 8.1.

jvanschie:
@Ocramius any progress made? really need it for a 8.1 project

Ocramius:
Nope: working on this component gives me massive anxiety (not sure if that is the exact naming - haven't gotten a professional to diagnose it) these days, so I'm probably holding off until I feel better about it.

bc-anton-k:
> Nope: working on this component gives me massive anxiety (not sure if that is the exact naming - haven't gotten a professional to diagnose it) these days, so I'm probably holding off until I feel better about it.

Hello @Ocramius,

Don't want to stress and/or pressure but maybe there are any updates for this PR?

TomA-R:
hey @Ocramius, is there a way forward with this? Is there something we can help with to get this to a state you're happy with? PHP 8.0 is now out of active support and many people are looking at upgrading before EOL later this year, so getting some resolution would be really appreciated. Thanks in advance!

TomA-R:
For anyone watching this issue, we've swapped this library for https://github.com/FriendsOfPHP/proxy-manager-lts. Considering PHP 8.0 no longer has active support and will discontinue security support in 5 months, it seems like this library should no longer be used.

Ocramius:
FWIW, I stopped most maintenance activities except security maintenance here precisely because of the LTS  fork butting in: this library is now a source of huge anxiety and frustration for me.

I may pick it up again if my brain decides that it isn't a source of sadness (or my therapist finds some magic), instead of pride and joy as it used to be, but I fully understand that people just want to get stuff done.

It's almost an exemplary case of MIT licensing being a terrible choice: should've been AGPL+Commercial from the start, but I was young and foolish.

PR #739, opened 2022-03-01
Generate tentative types when appropriate

nicolas-grekas:
I'm pretty sure this is not the approach you'd like to merge, but here is one that works, in case it helps a bit.

Branch 2.15.x is still for php `~8.0.0`. I don't know how you want to handle support for PHP 8.1.

I can update once all this is clarified.

### DIFF
--- a/src/ProxyManager/Generator/MethodGenerator.php
+++ b/src/ProxyManager/Generator/MethodGenerator.php
@@ -4,15 +4,20 @@ declare(strict_types=1);
 
 namespace ProxyManager\Generator;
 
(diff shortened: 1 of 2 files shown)


Ocramius:
Generating the `[#ReturnTypeWillChange]` blocks is probably the healthiest approach, since it doesn't break reflection for downstream consumers of proxies (which may probe the proxy API)

Ocramius:
Oof, sorry, wasn't an intentional close: fat-fingered that.

nicolas-grekas:
Merged into #738, which is the MVP (minimum viable patch) to move to PHP 8.1.

PR #763, opened 2022-07-11
Allow PHP 8.1

### LINKED ISSUE #721: Add PHP 8.1 to CI pipeline

Fredthelead:
- Changed the PHP `~8.0.0` constraint to `~8.0`
- Ran composer update with composer `2.3.9` and PHP `8.0.21`
- Added PHP `8.1` to the github pipes
- Updated psalm config and suppressed a false-positive

Fixes https://github.com/Ocramius/ProxyManager/issues/721

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="4.24.0@06dd975cb55d36af80f242561738f16c5f58264f">
+  <file src="src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php">
+    <MissingFile occurrences="1">
+      <code>require $fileName</code>
+    </MissingFile>
+  </file>
+</files>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,8 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.24.0@06dd975cb55d36af80f242561738f16c5f58264f">
-  <file src="src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php">
-    <MissingFile occurrences="1">
-      <code>require $fileName</code>
-    </MissingFile>
-  </file>
-</files>
--- a/psalm.xml
+++ b/psalm.xml
@@ -3,8 +3,8 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
-    totallyTyped="true"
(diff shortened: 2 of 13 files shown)


psalm.xml
```diff
@@ -3,8 +3,8 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
-    totallyTyped="true"
     rememberPropertyAssignmentsAfterCall="false"
+    errorBaseline="psalm-baseline.xml"
```
Ocramius: This shouldn't exist on this library 🤔

psalm-baseline.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="4.24.0@06dd975cb55d36af80f242561738f16c5f58264f">
+  <file src="src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php">
+    <MissingFile occurrences="1">
+      <code>require $fileName</code>
```
Ocramius: This should be suppressed inline: it is an intentional design, not an issue that is to be fixed at some point 🤔

Ocramius (changes requested):

psalm.xml
```diff
@@ -3,7 +3,6 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
-    totallyTyped="true"
```
Ocramius, line 6: I think this requires `errorLevel="1"` as replacement 🤔

Fredthelead: My bad 🤦

dfreudenberger:
Anything that needs to be done in order to get this merged & released? Happy to help if something is missing.

Ocramius:
Release requires this work: https://github.com/Ocramius/ProxyManager/milestone/56

Fredthelead:
The problem with this PR is that the tests won't work with locked dependencies on both php 8.0 and 8.1. Not sure how to solve this without removing some tests.

Ocramius:
Supporting `8.0` and `8.1` at the same time might indeed be challenging.

I can suggest skipping some tests on `8.0`, so we focus on releasing a version that works on both PHP `8.0` and `8.1`, then drop `8.0` support in a new minor, which in turn removes all conditionals that disable some tests.

jakublabno:
Hi, any update on this?

Ocramius:
https://github.com/Ocramius/ProxyManager/pull/738#issuecomment-1296944093

PR #765, opened 2022-10-27
Remove codelicia/xulieta from the dev dependencies list

malukenho:


Ocramius (approved):
Thanks @malukenho

The remaining failures are up to me here.

PR #780, opened 2023-03-17
Make property private in final class

lstrojny:
This should probably be private since it resides in a final class.


Ocramius (approved):
Thanks @lstrojny!

PR #785, opened 2023-05-16
Promoted constructor properties should not be doubled, fix #784

### LINKED ISSUE #784: Nullable constructor property promotion support
Nullable modifier is lost in promoted constructor properties. For example, this code will result fatal error:
 
```php
class Test {
    public function __construct(
        protected ?int $value
    ) {
    }
}

$factory = new \ProxyManager\Factory\LazyLoadingValueHolderFactory();

$proxy = $factory->createProxy(
    Test::class,
    function (& $wrappedObject, $proxy, $method, $parameters, & $initializer) {
        $wrappedObject = new Test(null);
    }
);
```

```
PHP Fatal error:  Type of ProxyManagerGeneratedProxy\__PM__\Test\Generated1c12fa522c43907846cdf9382a0d356f::$value must be ?int (as in class Test) in .../vendor/ocramius/proxy-manager/src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php(54) : eval()'d code on line 3
```

And indeed, property is no longer nullable in generated proxy:
```php
namespace ProxyManagerGeneratedProxy\__PM__\Test;

class Generated7ef5eb7812f288337a0d82be0b42baa6 extends \Test implements \ProxyManager\Proxy\VirtualProxyInterface
{
...

    public function __construct(protected int $value) // <---------------- nullable modificator is lost
    {
        static $reflection;

        if (! $this->valueHoldere9785) {
            $reflection = $reflection ?? new \ReflectionClass('Test');
            $this->valueHoldere9785 = $reflection->newInstanceWithoutConstructor();
        unset($this->value);

        }

        $this->valueHoldere9785->__construct($value);
    }
...
```

Grundik:
Since properties are inherited, their declaration should not be doubled.

Issue arises with laminas-code versions, which supports constructor property promotion (at least 4.7, maybe earlier).

Closes #784 

// cc @nicolas-grekas

### DIFF
--- a/src/ProxyManager/Generator/MethodGenerator.php
+++ b/src/ProxyManager/Generator/MethodGenerator.php
@@ -6,6 +6,7 @@ namespace ProxyManager\Generator;
 
 use Laminas\Code\Generator\DocBlockGenerator;
 use Laminas\Code\Generator\MethodGenerator as LaminasMethodGenerator;
+use Laminas\Code\Generator\ParameterGenerator;
 use Laminas\Code\Reflection\MethodReflection;
 
 /**
@@ -19,10 +20,23 @@ class MethodGenerator extends LaminasMethodGenerator
     public static function fromReflectionWithoutBodyAndDocBlock(MethodReflection $reflectionMethod): self
     {
         /** @var static $method */
-        $method = parent::copyMethodSignature($reflectionMethod);
+        $method = static::copyMethodSignature($reflectionMethod);
 
         $method->setInterface(false);
         $method->setBody('');
+
+        return $method;
+    }
+
+    public static function copyMethodSignature(MethodReflection $reflectionMethod): parent
+    {
+        $method = parent::copyMethodSignature($reflectionMethod);
+
+        foreach ($reflectionMethod->getParameters() as $reflectionParameter) {
+            $method->setParameter(
+                ParameterGenerator::fromReflection($reflectionParameter)
+            );
+        }
 
         return $method;
     }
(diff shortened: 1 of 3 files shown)


src/ProxyManager/Generator/MethodGenerator.php
```diff
@@ -19,14 +20,27 @@ class MethodGenerator extends LaminasMethodGenerator
...
+    {
+        $method = parent::copyMethodSignature($reflectionMethod);
+
+        foreach ($reflectionMethod->getParameters() as $reflectionParameter) {
+            $method->setParameter(
+                ParameterGenerator::fromReflection($reflectionParameter)
+            );
+        }
+
+        return $method;
```
Ocramius, line 41: Shouldn't this behavior be fixed in `laminas/laminas-code`, if we're required to copy parameter declarations twice to fix it?

Ocramius (changes requested):

src/ProxyManager/Generator/MethodGenerator.php
Grundik: I think laminas-code does its best: it detects property promotion, and copies that as requested. But since we are doing inheritance, we dont need that. So problem on usage side, not on provider one.

Solution for laminas-code would be just to drop support of that feature of PHP, which is not the best, I think.

Grundik: There might be the another problem in laminas-code, since nullable flag are dropped in these promoted properties. But thats another issue. For doublers there should not be constructor property declarations anyways: they are already declared in base class.

Grundik: I've created another issue for this (separate!) problem: https://github.com/laminas/laminas-code/issues/183

